Issue #8, opened 2017-03-09
Add a global flag to ELASTICSEARCH_DSL setting which disables automatic indexing


markotibold:
Hey,

In some cases it can be useful to disable the automatic indexing, for example when loading lots of fixtures. Or in our case when testing the success of fixture-loading in a CI environment with no ES backend.

Would you appreciate a pull request that implements such a flag?

sabricot:
Hi,

Yes it's a good idea, if you want to implement it it will be with pleasure.

markotibold:
Cool,

I'll come up with something soon.

PR #10, opened 2017-03-15
Implements #8 by adding a global settings flag for disabling autosyncing

markotibold:


sabricot:
Thanks for your contribution !

PR #12, opened 2017-03-28
Elasticsearch5

markotibold:
Hey,

This pull request solves compatibility issues with elasticsearch5 that I am aware of.

For example the helper function `_make_dsl_class` was removed. so I've back-ported it into this project. It's of course not the most elegant solution. But it works.

Don't hesitate to reject the PR if it's not elegant enough. Maybe we can discuss a better solution here.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,7 +24,7 @@ Features
 - Requirements
    - Django >= 1.8
    - Python 2.7, 3.4, 3.5
-   - Elasticsearch >= 2.1
+   - Elasticsearch >= 2.1,<=5.1.0
 
 .. _Search: http://elasticsearch-dsl.readthedocs.io/en/stable/search_dsl.html
 
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -2,19 +2,40 @@ import collections
 from types import MethodType
 
 from django.db import models
-
 from elasticsearch_dsl.field import (
     Boolean,
     Date,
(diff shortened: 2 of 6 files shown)


safwanrahman:
Is it possible to escape using the internal function ``_make_dsl_class``?
Just a thought, nothing serious.

markotibold:
Hi, I'm not really sure I understand your point, however the function _make_dsl_class was removed from elasticsearch-dsl-py in this commit https://github.com/elastic/elasticsearch-dsl-py/commit/8d177de0ddea43b9ed608b2c66f300ee541715d0

safwanrahman:
@markotibold I was thinking if its possible to rewrite it somehow so we should no longer use ``_make_dsl_class`` or we can use another function that is not ``internal``. Be mentioned that, function starts with ``_`` are treated as internal function and for internal use. so the contributor of ``elasticsearch-dsl`` can remove it without any deprecation warning.

barseghyanartur:
@safwanrahman, @sabricot:

I think this pull request is just fine. Making sure nothing breaks for older versions and assuring the compatibility with feature versions. Merge would help further development of the package.

sabricot:
Thank you for the PR and sorry for the lag.
I promise i will dig into it this weekend :)
Just an observation for now, i don't think it handles the new fields added in ES5 (text/keyword...)

barseghyanartur:
@sabricot: 

Sorry for bugging, but is there any progress on this?

PR #16, opened 2017-05-01
Elasticsearch 5 compatibility with new fields

HansAdema:
A continuation of the work from @markotibold , this PR includes not separate subclasses for fields rather than relying on the previously removed `_make_dsl_class` method. It also includes the TextField and KeywordField types for ES 5 compatibility.

### DIFF
--- a/django_elasticsearch_dsl/__init__.py
+++ b/django_elasticsearch_dsl/__init__.py
@@ -4,18 +4,25 @@ from .documents import DocType  # noqa
 from .indices import Index  # noqa
 from .signals import delete_document, update_document # noqa
 from .fields import (  # noqa
-    StringField,
-    FloatField,
-    DoubleField,
-    ByteField,
-    ShortField,
-    IntegerField,
-    LongField,
-    DateField,
+    AttachmentField,
     BooleanField,
-    ObjectField,
-    NestedField,
+    ByteField,
+    CompletionField,
+    DateField,
+    DoubleField,
+    FloatField,
+    GeoPointField,
+    GeoShapeField,
+    IntegerField,
+    IpField,
+    KeywordField,
     ListField,
+    LongField,
+    NestedField,
+    ObjectField,
+    ShortField,
+    StringField,
+    TextField,
 )
 
 __version__ = '0.1.0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,27 +5,42 @@ language: python
 python:
   - "3.6"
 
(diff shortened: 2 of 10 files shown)


django_elasticsearch_dsl/__init__.py
```diff
@@ -4,18 +4,25 @@
...
     LongField,
-    DateField,
-    BooleanField,
-    ObjectField,
     NestedField,
-    ListField,
+    ObjectField,
+    ShortField,
+    StringField,
+    TextField,
```
sabricot: It is not backward compatible with `elasticsearch-dsl>=2.0.0,<3.0.0`. Maybe we can just add a try catch for the TextField and KeywordField

HansAdema:
I forgot about the ES 2 compatibility.

I added separate envs for ES 2 and ES 5 to tox and separate APT urls to the Travis config and confirmed what you already said: this branch doesn't work with Elasticsearch 2.

If you have any ideas on how to support ES 2 and ES 5 simultaneously, please do implement it.

drdaeman:
Hi! I'm watching this PR progress and have a suggestion. How about moving all ES5-specific fields in a separate module?

I mean:

1. Rename `fields.py` to `fields/base.py`. Use pre-ES5 version, with only ES2-specific stuff.

2. Add `fields/es5.py` with ES5-specific stuff

       from elasticsearch_dsl.field import (
           Attachment,
           Byte,
           ...
       )
       from .base import DEDField
   
       class AttachmentField(DEDField, Attachment):
           pass
       ...

3. And in `fields/__init__.py` do something like this:

       from .base import *
       try:
           from .es5 import *
       except ImportError:
           pass

What do you think?

HansAdema:
@drdaeman That's an implementation I can get behind, separating the modules would help prevent a large part of the ImportError mess you get when trying to support multiple versions.

How about having separate branches for the major Elasticsearch versions? The same approach is taken by elasticsearch-dsl as well, and I think mirroring the branches and versions in Elasticsearch, DSL and DED would be good.

drdaeman:
Better ask @sabricot's opinion on this. I'm just a random stranger passing by ;)

Personally, though, I think as long as the differences are minimal, multiple branches would only unnecessarily complicate things. Maintainers mileage may vary, but just a bunch of extra classes that appear in ES5 but not ES2 doesn't look like a big deal to me and probably not worth the hassle of maintaining multiple versions. Of course, that's only as long as all the differences are just those few extra classes - but if nothing else is currently planned, branching can be postponed.

HansAdema:
Fixed with PR #21

PR #17, opened 2017-05-04
Refactor search_index command

HansAdema:
I got annoyed having to type `y` many times when running the delete command, so I refactored it to ask to delete all indexes at the same time.

I also changed it so the rebuild command exists if the delete is not confirmed (because you'd expect it to cancel the rebuild if you say no).


sabricot:
Thank you for your contribution, i merged it in #19

HansAdema:
Looking good! I'm closing this PR because the issue has been resolved.

PR #23, opened 2017-06-02
Support defining DocType indexes through Meta class

HansAdema:
Currently, DocType classes must be assigned to indexes using decorators. When setting the index property in the Meta class, the DocType will not be added to the registry.

This change makes it possible to do that.

It did require me to add additional checks to the registry to prevent the same name from being used multiple times.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -12,4 +12,4 @@ Contributors
 ------------
 
 markotibold
-HansAdema
+HansAdema - Devhouse Spindle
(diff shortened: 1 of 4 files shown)


HansAdema:
Actually, hold off on merging this, I just found out it doesn't properly pick up the field settings (analyzers and stuff) using this notation.

HansAdema:
OK, fixed now.

This change ensures you can define the index of your doc type in the meta class. It will break if you define the index twice, but so does Elasticsearch DSL itself. You can still use the decorator in case that happens.

sabricot:
Thank you, it's a nice add

PR #28, opened 2017-06-20
Add options to disable automatic index refreshes.

### LINKED ISSUE #26: Add option whether always to refresh
Currently, Django Elasticsearch DSL forces an index refresh on every update. While this ensures data is available for search immediately, it also generates a high server load in a write intensive environment, and is just wasting good server power if being able to search that fast is not a priority.

While the `DocType.update` method has an argument `refresh`, most refreshes are triggered by signals which does not support passing that option.

An global option and/or an option per doc type (meta class?) would help prevent unnecessary load on the Elasticsearch servers.

HansAdema:
With this change, you now have more options to disable automatic refreshes, in order of preference:

1. The `refresh` argument passed to the `DocType.update()` method.
2. The `auto_refresh` attribute defined in the DocType Meta class.
3. The `ELASTICSEARCH_DSL_AUTO_REFRESH` Django setting.
4. The default `True` value (for backwards compatibility).

Previously, it wasn't possible to prevent forced refreshes using the signals.

Fixes #26

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,6 +107,7 @@ define in a ``documents.py`` file.
             ] # the fields of the model you want to be indexed in Elasticsearch
 
             # ignore_signals = True # To ignore auto updating of Elasticsearch when a model is save or delete
+            # auto_refresh = False  # Don't perform an index refresh after every update (overrides global setting)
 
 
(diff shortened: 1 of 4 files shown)


sabricot:
Another good addition. Thank you, i merged it.

PR #29, opened 2017-06-20
Remove `iterator()` from `search_index._populate()`

### LINKED ISSUE #27: search_index populate ignores prefetch_related
In most of my DocType classes, I extended `get_queryset` to add select_related and prefetch_related clauses to speed up indexing. The select_related clauses work fine, but since populate takes the iterator of the queryset, the prefetch_related clauses are ignored.

To help speed up indexing, it would help a lot of an alternative way to process the models could be used (like using pagination?).

Source: https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.iterator

HansAdema:
The use of `iterator()` in `search_index._populate()` breaks the usage of `prefetch_related`, which is useful to speed up indexing. I have not been able to find any advantage in using `iterator()`.

Fixes #27

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -116,8 +116,7 @@ class DocType(DSLDocType):
         """
         Return the queryset that should be indexed by this doc type.
         """
-        qs = self._doc_type.model._default_manager
-        return qs
+        return self._doc_type.model._default_manager.all()
 
     def prepare(self, instance):
         """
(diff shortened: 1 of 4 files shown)


diwu1989:
`I have not been able to find any advantage in using iterator().`
If you're working with large scale production data, and using `prefetch_related`, not using `iterator()` is going to make indexing jobs impossible.

HansAdema:
@diwu1989 Can you please explain more about what you mean by that? In my experience, adding prefetch_related clauses drastically helps speed up indexing. The results are paginated, so IMO it works fine with large tables as well.

sabricot:
One advantage could be the server-side cursor from django 1.11 https://docs.djangoproject.com/en/1.11/ref/databases/#postgresql-server-side-cursors but maybe we could add a simple pagination system controlled by a settings to replace the iterator or just give an example of how do it.

sabricot:
I merged it in https://github.com/sabricot/django-elasticsearch-dsl/pull/50. Thank you!

PR #30, opened 2017-06-26
Replace simple model signals with easier to customise signal processors

barseghyanartur:
I have been recently busy implementing an app to integrate Elasticsearch DSL with Django REST framework.

It turned into a [django-elasticsearch-dsl-drf](https://github.com/barseghyanartur/django-elasticsearch-dsl-drf) package, which make use of [django-elasticsearch-dsl](https://github.com/sabricot/django-elasticsearch-dsl/).

I want to avoid heavy signals in real time, but rather would add them to the queue to be processed later on.

Therefore - the introduction of signal processors.

At the moment, functionality remains intact, but leaves a place for handy replacement of real time index updates with custom ones (for instance, delayed update using Celery).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -474,6 +474,22 @@ Default: ``True``
 
 Set to ``False`` to globally disable auto-syncing.
 
+ELASTICSEARCH_DSL_SIGNAL_PROCESSOR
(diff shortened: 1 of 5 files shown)


sabricot:
Thank you it's a great addition ! I just merged it.

PR #33, opened 2017-06-30
Add support for resynching ES documents if related models are updated

HansAdema:
If you include data from related models in your Elasticsearch documents, you will probably need to update the documents whenever the related models are updated as well. Previously, this did not happen automatically.

This change allows you to automatically update an Elasticsearch document if a related model is updated. Simply add a list of model classes to the Meta class as `related_models` to update, and define a `get_instances_from_related` method to map the related models to models of the doc type to update.

I'm not sure about the mapping method part and I would appreciate some thoughts on that, but otherwise I think this is a useful addition.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -296,12 +296,17 @@ You can use an ObjectField or NestedField.
             'pk': fields.IntegerField(),
         })
 
+        def get_instances_from_related(self, instance):
+            """If related_models is set, define how to retrieve the Car instances from the related model."""
+            return instance.car_set.all()
+
         class Meta:
             model = Car
             fields = [
                 'name',
                 'color',
             ]
+            related_models = [Manufacturer]  # Optional: to ensure the Car will be re-saved when Manufacturer is updated
 
         # Not mandatory but to improve performance we can select related in
         # one sql request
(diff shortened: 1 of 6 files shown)


barseghyanartur:
Top, man!

diwu1989:
Let's say that u have models A, B, C
A relates to B, B relates C, but A does not relate to C directly
If an instance of A is saved, the related instance of B is also updated.
But does the instance of C also get updated?

HansAdema:
@diwu1989 I don't fully what you mean by "instance" and A, B and C in this regard, but I can try to explain.

Suppose you have a model Car, Manufacturer and Country. Cars are created by a Manufacturer and Manufacturers are located in a certain country.

If the document of Car contains just the attributes of the Car model, then you don't need to define related models. If the document of Car contains a nested object with the information of Manufacturer, then you should define Manufacturer as a related model for Car. That way, if a Manufacturer is updated, any Cars of that manufacturer will be reindexed in Elasticsearch.

If you also want to include Country information into the car document (as a sub object of Manufacturer), then you also have to specify Country as a related model to Car. Relations are not traversed, specifying it as a related model to Manufacturer only affects the Manufacturer document and not the Car document.

So as a rule of thumb: you would specify all the models whose data is included in the doc type.

sabricot:
Thanks for you great work 👍

PR #58, opened 2017-10-25
Filter populate command

sajadbahar:
Here is implement of populate with filter,

you could use Django filter phrase like `{field_name}__gte`, ...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -482,6 +482,12 @@ Populate the Elasticsearch mappings with the django models data (index need to b
 
     $ search_index --populate [--models [app[.model] app[.model] ...]]
 
+Populate with query filter, also you could pass django style filter to populate (index need to be existing):
+
+::
+
+    $ search_index --populate [--models [app[.model] app[.model] ...] [--limit [[field value] [field__style value] ...]]
+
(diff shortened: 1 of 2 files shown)


herrnici:
Any progress on fixing and merging this? I find this an important feature.

safwanrahman:
@herrnici Can you explain your usecase?

herrnici:
@safwanrahman A large table of versioned data (division into multiple models is not an option), where only current data is needed for search, and therefore only current data should be indexed. Of course this would also work by

- filtering every search (wastes space, slight runtime impact)
- manually deleting the majority of the data from the index after every build (inconvenient, temporarily wastes space), or
- changing the default manager to only include partial data (has impacts on other components).

But I feel needing only a subset of data isn't that uncommon a use case.

alexgarel:
@herrnici, given your usecase, why don't you simply override `get_queryset` in you `Document`  class ?

herrnici:
Good point, that should cover all cases. No idea why I didn't think of that.
In that case this PR is really somewhat redundant. Thank you and sorry for the ping.

PR #62, opened 2017-11-15
Strings are dead, long live strings!

stevelacey:
Elasticsearch 5.0 removes the string type, replacing them with text and keyword types.

https://www.elastic.co/blog/strings-are-dead-long-live-strings

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -361,7 +361,7 @@ So for example you can use a custom analyzer_:
     class CarDocument(DocType):
         description = fields.TextField(
             analyzer=html_strip,
-            fields={'raw': fields.StringField(index='not_analyzed')}
+            fields={'raw': fields.KeywordField()}
         )
 
         class Meta:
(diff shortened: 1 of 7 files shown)


stevelacey:
I see you do actually already support but `model_field_class_to_field_class` still defaults to using StringField's... so I'll close this for now, but maybe for ES5 the defaults should be to TextField's?

sabricot:
Yes you are right i should probably add a settings for this.

stevelacey:
I am doing this for the time being:

```py
from django_elasticsearch_dsl import documents, fields

# monkey-patch model_field_class_to_field_class mapping so that we use
# Elasticsearch 5.0 text fields instead of deprecated string fields
documents.model_field_class_to_field_class.update({
    documents.models.CharField: fields.TextField,
    documents.models.EmailField: fields.TextField,
    documents.models.FilePathField: fields.TextField,
    documents.models.SlugField: fields.TextField,
    documents.models.TextField: fields.TextField,
    documents.models.URLField: fields.TextField,
})
```

Issue #15, opened 2017-04-16
Auto-indexing for nested fields


yarnball:
Hi,

When I update a nested field value, the change is not being indexed in elastic search.

I have followed your example, but had no success with making the "Tag" become indexed when its value changes.

Could you please give an example of how this is done?

```
class Car(models.Model):
    name = models.CharField()
    manufacturer = models.ManyToMany('Tag')

class Tag(models.Model):
    name = models.CharField()
    def tags(self):
        return self.tag_set.all()
```

```
car = Index('cars')


@car.doc_type
class CarDocument(DocType):

    tag = NestedField(properties={
        'name': StringField(),
    })

    class Meta:
        model = Car
        fields = [
            'name',
        ] 
```

alexislg2:
I guess this is the normal behaviour. But I have the same problem: I want that a save on the nested object triggers a refresh

sabricot:
There is now a way to do this:

```
class Car(models.Model):
    name = models.CharField()
    tags = models.ManyToMany('Tag')

class Tag(models.Model):
    name = models.CharField()
```

```
car = Index('cars')


@car.doc_type
class CarDocument(DocType):

    tags = NestedField(properties={
        'name': StringField(),
    })

    class Meta:
        model = Car
        related_models = [Tag]
        fields = [
            'name',
        ]

    def get_instances_from_related(self, tag_instance):
        return tag_instance.car_set.all()
```

palqcho:
Hi All,

Did anyone make it work?
It does not work at our side and we have tried literally everything :(
Here is our code:

models.py:

class ExpertTypes(models.Model):
    name = models.CharField(max_length=100)

    def __unicode__(self):
        return self.name


class Experts(models.Model):
    expert_type = models.ManyToManyField('ExpertTypes')
    ml_rank = models.ForeignKey(ExpertMlRankStatus, null=True)
    name = models.CharField(max_length=100)


documents.py:

experts_index = Index('experts-index')


@experts_index.doc_type
class ExpertsIndexDoc(DocType):
    expert_type = fields.NestedField(properties={
        'name': fields.StringField()
    })

    ml_rank = fields.ObjectField(properties={
        'ml_rank': fields.StringField()
    })

    class Meta:
        model = Experts
        # Optional: to ensure the Experts will be re-saved when ExpertTypes and ExpertMlRankStatus are updated
        related_models = [ExpertTypes, ExpertMlRankStatus]
        fields = [
            'name',
[...]

PR #63, opened 2017-11-20
Now handle m2m signals for auto syncing

sabricot:
Should fix the issues of auto sync when adding or removing
object to a m2m relationship.
Related to #15


palqcho:
I tested the branch and I confirm that the m2m relationship issue have been fixed. Also, no regressions have been found.

PR #73, opened 2017-12-14
Implement ActionBuffer to speed up runtime indexing

HansAdema:
One notable issue I have with how the DED synchronization logic works, is that it sends a a call to Elasticsearch for every single document instance. If you have a complex set of relations in a model and a signal is being triggered, it will cause many separate Elasticsearch calls which can give quite a hit to page load times.

Instead, I built `ActionBuffer` to easily combine indexing actions and send them away in a single bulk query per connection/cluster.

There are two ways to add actions to the buffer. You could use:
- `add_doc_actions`: Synchronize model(s) for a given DocType.
- `add_model_actions`: Synchronize all DocTypes for a model or iterable.

This results in a number of changes:
- Data persistence and synchronization is removed from the registry and contained in `ActionBuffer`. The registry only handles registering stuff now. `DocumentRegistry.update` and `DocumentRegistry.delete` are only kept for backwards compatibility.
- Signal specific logic is contained in the signals. Calling `DocumentRegistry.update` manually outside signals will not longer check signal flags.
- Signal performance should increase substantially for complex relation structures, as only one more ES call will be sent per signal call.
- You can directly use `ActionBuffer` to build your own, efficient model synchronization logic (any takers for Django Rest Framework integration?).

### DIFF
--- a/django_elasticsearch_dsl/actions.py
+++ b/django_elasticsearch_dsl/actions.py
@@ -1,0 +1,115 @@
+from itertools import chain
+
+from collections import defaultdict
+
+from django.core.paginator import Paginator
+from django.db.models import Model
+from elasticsearch.helpers import bulk
+from six import iteritems
+
+from .apps import DEDConfig
+
+
+class ActionBuffer(object):
+    """
+    ActionBuffer helps to gather index/delete actions and send them in bulk.
+    """
+
+    def __init__(self, registry=None):
+        if registry:
+            self._registry = registry
+        else:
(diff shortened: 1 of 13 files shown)


andreyrusanov:
Any updates on this one? Looks like nice feature to have.

sabricot:
Yes i am agree, I'am gonna merge it in next weeks.

sabricot:
Closing in favor of #147 thanks again

Issue #79, opened 2018-01-06
TypeError on new ArrayField with document ListField 


CBinyenya:
I Just added a new postgres array field to a table and specified the document field as a list
`contacts = ArrayField(models.CharField(max_length=13, blank=True), size=3, null=True)`

and 
`'contacts': fields.ListField(fields.StringField()),`

while rebuilding the index I get  `TypeError: 'NoneType' object is not iterable` error

PR #80, opened 2018-01-06
Added check for list field to avoid iterating over None types from Ar…

CBinyenya:
#79 …ryFields with null = True


sabricot:
Thank you for your contribution !

Issue #34, opened 2017-07-02
Implement CelerySignalProcessor


barseghyanartur:
Now that Signal Processors are merged into master, it would be handy to have a built-in Celery signal processor. 

I'm gonna work on it, if you don't mind, @sabricot.

sabricot:
Yes it would be nice :)

MightySCollins:
@barseghyanartur did you manage to make any progress on this? If not I might have to try something myself.

barseghyanartur:
@MightySCollins:

Not yet. I'll probably postpone it for a couple of weeks. If could spend time on this now, go on, please.

hemil:
Any updates on this?
@barseghyanartur

barseghyanartur:
@hemil:

No, unfortunately not, but that's on my list.

CorrosiveKid:
Check out #87. It is currently working with Celery 4, but need to write tests around the new signal processor, and also need to add some README/documentation updates.

hemil:
Alright thanks, will check it out

janwout:
@safwanrahman @barseghyanartur @CorrosiveKid what is needed for a celery signal processor to be added to the project?

andreasnuesslein:
@janwout I'd think a PR would help ;)

I started to build this a while back, but ran into some "logic" problems because you can't pass along instances to celery tasks (as you probably know):

```python
@shared_task
def task_handle_save(pk, content_type):
    ct_model = ContentType.objects.get_by_natural_key(*content_type)
    instance = ct_model.get_object_for_this_type(pk=pk)
    registry.update(instance)
    registry.update_related(instance)

class CelerySignalProcessor(RealTimeSignalProcessor):
    def handle_save(self, sender, instance, **kwargs):
        ct = ContentType.objects.get_for_model(instance)
        task_handle_save.delay(instance.pk, ct.natural_key())
```

So basically what I did was to just move the `registry` logic into a task. The I get the `content_type` string and the instances `id` and pass it to that celery task.
Now the problem is: As soon as you do that for deletion, Celery won't find your instance anymore because it will already have been deleted. 

Maybe this helps you to get started though :)

alexgarel:
Yes the good way of handling this with celery is to pass the id of the database object to save.

But there is another pitfall:  if your database is running transactions, you have to defer the task until the commit. Otherwise, you may have a race condition, where celery get the object while the transaction is not finished and thus index an old version of the object.
[...]

PR #87, opened 2018-03-08
Implement the Celery signal processor

CorrosiveKid:
The `CelerySignalProcessor` allows automatic updates on the index as delayed background tasks using Celery.

NB: We cannot process deletes as background tasks. By the time the Celery worker would pick up the delete job, the model instance would already deleted. We can get around this by setting Celery to use `pickle` and sending the object to the worker, but using `pickle` opens the application up to security concerns.

#34

### DIFF
--- a/django_elasticsearch_dsl/signals.py
+++ b/django_elasticsearch_dsl/signals.py
@@ -6,6 +6,7 @@ cause things to index.
 
 from __future__ import absolute_import
 
+from celery.exceptions import ImproperlyConfigured
 from django.db import models
 from django.apps import apps
 
@@ -73,13 +74,10 @@ class BaseSignalProcessor(object):
         registry.delete(instance, raise_on_error=False)
 
 
-class RealTimeSignalProcessor(BaseSignalProcessor):
-    """Real-time signal processor.
-
-    Allows for observing when saves/deletes fire and automatically updates the
-    search engine appropriately.
+class DjangoSignalsMixin(object):
     """
-
+    Enable Django signals integration
+    """
     def setup(self):
         # Listen to all model saves.
         models.signals.post_save.connect(self.handle_save)
@@ -96,12 +94,21 @@ class RealTimeSignalProcessor(BaseSignalProcessor):
         models.signals.m2m_changed.disconnect(self.handle_m2m_changed)
         models.signals.pre_delete.disconnect(self.handle_pre_delete)
 
+
+class RealTimeSignalProcessor(DjangoSignalsMixin, BaseSignalProcessor):
+    """Real-time signal processor.
+
+    Allows for observing when saves/deletes fire and automatically updates the
+    search engine appropriately.
+    """
+    pass
+
 try:
     from celery import shared_task
 except ImportError:
     pass
 else:
-    class CelerySignalProcessor(RealTimeSignalProcessor):
+    class CelerySignalProcessor(DjangoSignalsMixin, BaseSignalProcessor):
         """Celery signal processor.
 
         Allows automatic updates on the index as delayed background tasks using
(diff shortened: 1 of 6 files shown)


CorrosiveKid:
This is still work in progress. Need help writing the tests around the new signal processor.

django_elasticsearch_dsl/signals.py
```diff
@@ -94,3 +95,58 @@ def teardown(self):
...
+
+            Given an individual model instance, update the object in the index.
+            Update the related objects either.
+            """
+            pk = instance.pk
+            app_label = instance._meta.app_label
+            model_name = instance.__class__.__name__
+
+            self.registry_update_task.delay(pk, app_label, model_name)
+            self.registry_update_related_task.delay(pk, app_label, model_name)
```
ur001: ### Use `Concrete model` to work correctly with `.only()` / `.defer()`
```python
user = User.objects.only('email')[0]
>>> user.__class__.__name__
'User_Deferred_account_name_avatar_avatar150_ua1ae4ad6a3bf5a2fb7a2d93a6925040a'
>>> user._meta.concrete_model.__name__
'ExpUser'
```

### Check model in registry, to avoid unnecessary tasks
```python
pk = instance.pk
model = instance._meta.concrete_model
model_name = model.__name__
app_label = instance._meta.app_label

if model in registry._models:
    self.registry_update_task.delay(pk, app_label, model_name)

if model in registry._related_models:
    self.registry_update_related_task.delay(pk, app_label, model_name)
```

andreyrusanov:
@CorrosiveKid do you still need help with tests?

CorrosiveKid:
@andreyrusanov Yes please, if possible. I still haven't had the time to look at it after my initial commit.

django_elasticsearch_dsl/signals.py
CorrosiveKid: Thanks! I'll make the changes when I have some time.

CorrosiveKid:
@andreyrusanov Done!

andreyrusanov:
@ur001 @CorrosiveKid @sabricot 

we have an issue with tests here. Signals could not be properly attached to models, used for tests. They connected, but don't triggered. What I've tried:
- I used override_settings decorator (with it old signal were triggered anyway, unless I removed it explicitly)
- tried to run DEDConfig().ready() again (with signal_processor field set to None)
- tried to explicitly disable old signals and connect new one.
I also tried to do lots of minor changes/things, but nothing works.

New signals were connected without issue, but when afterwards something happened with model Signal dispatcher doesn't find any receiver for it.. If anybody can take a look or have an idea how to fix it - please do!

I commited workaround I've made. It is really hacky in some point and I have really big doubts it will be approved. If so - I or @CorrosiveKid can just revert it back. If not - I will resolve conflicts/add extra docs.

gkapatia:
any update on this branch?

divick:
@CorrosiveKid I see there is one issue with the current implementation. It calls a task for every save, while it could possibly call a task for multiple objects in one go using the elasticsearch_dls.helpers.bulk method. It is not simple though to implement this behaviour as it will need to then have a queue which stores objects of same types in different queues and then passed at certain interval to bulk update.

emilio-rst:
@CorrosiveKid I suggest to use [transaction.on_commit](https://docs.djangoproject.com/en/2.2/topics/db/transactions/#performing-actions-after-commit) to prevent race condition.

`transaction.on_commit(lambda: self.registry_update_task.delay(pk, app_label, model_name))
transaction.on_commit(lambda: self.registry_update_related_task.delay(pk, app_label, model_name))`

vonmaster:
Is this branch still in development? Do you need any help?

CorrosiveKid:
@vonmaster Any help is greatly appreciated, I haven't had the time to look at it yet.

ijharulislam:
@CorrosiveKid 
any update on this branch?

CorrosiveKid:
> @CorrosiveKid
> any update on this branch?

It's quite stale at this point since I haven't had the time to pursue this :(

patrickml:
@CorrosiveKid if you do ever end up finishing this you could handle deletes by getting the model class and creating a new instance of the model, but not saving it like `get_model(app_label, model_name)(pk=pk)` then padding that to the `registry.delete`

Andrew-Chen-Wang:
Quick check @safwanrahman: does this still work on the latest package version? Are there any breaking changes?

safwanrahman:
@Andrew-Chen-Wang This PR is actually out of date. Need to rebase against master and port it according to new code!

cr0mbly:
Would love to see this picked up and finished!

Bidaya0:
> Would love to see this picked up and finished!

Please take a look. 
https://github.com/django-es/django-elasticsearch-dsl/pull/414

I rebased it against master and implement the celery  delete  and delete_related tasks.

safwanrahman:
#414 is better position currently.

PR #92, opened 2018-04-13
Handle ObjectDoesNotExist for related instances

dsanders11:
Currently if `get_instances_from_related` raises an `RelatedObjectDoesNotExist` exception it escapes instead of being considered `related_model = None` like it should. This PR fixes that.

### DIFF
--- a/django_elasticsearch_dsl/registries.py
+++ b/django_elasticsearch_dsl/registries.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from itertools import chain
 
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.six import itervalues, iterkeys, iteritems
 
 from .apps import DEDConfig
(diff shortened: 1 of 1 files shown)


dsanders11:
PR updated to handle all invocations of `get_instances_from_related`.

phoebebright:
Can this request be merged?  Need this fix!

phoebebright:
Found also need to trap Attribute error::

         except ObjectDoesNotExist:
                related = None
         except AttributeError:
                related = None

This was running es 6, django2, python 3.6.1 with::
      elasticsearch==6.3.0
      elasticsearch-dsl==6.1.0

Note that the latest version of elasticsearch-dsl is 6.2.0 but this has removed DocType in init so this library currently requires <= 6.1.0

sabricot:
Thank you @dsanders11 and sorry for the delay again :)

PR #102, opened 2018-04-28
Fix perf when populate on huge qs

### LINKED ISSUE #86: Performance issue when populate millions documents
##### Problem
When we need to put a lot of documents in index, we need to use `queryset_pagination` meta option to paginate. Django pagination need a sorted queryset with `order_by` (cf doc)[https://docs.djangoproject.com/en/2.0/topics/pagination/#required-arguments] otherwise same pk can be present more than once.

Put `order_by` on queryset will make django paginator call order_by for each page. Call `order_by` on huge queryset (like 10 millions) will lead to a huge perfomance issue.

##### Temporary solution:

We can override `_get_actions` method (from `django_elasticsearch_dsl.documents.DocType`) to not use django paginator when a queryset is passed. More over because of the way a database index work, we should first fetch only pks, and then do sub request based on it.

```python
from django.db.models.query import QuerySet


def _get_actions(self, object_list, action):
    if self._doc_type.queryset_pagination and isinstance(object_list, QuerySet):
        pks = object_list.order_by('pk').values_list('pk', flat=True)
        for start_pk_index in range(0, len(pks), self._doc_type.queryset_pagination + 1):
            end_pk_index = start_pk_index + self._doc_type.queryset_pagination
            ranged_qs = object_list.filter(pk__range=[
                pks[start_pk_index],
                pks[end_pk_index]
            ])
            for object_instance in ranged_qs:
                yield self._prepare_action(object_instance, action)
    else:
        super()._get_actions(object_list, action)
```

Available to make the PR if needed.

ahivert:
Fix #86

- create a custom paginator based on pks
- Add test and refactor tests_documents to use models.py and documents.py
- make travis run flake8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,7 +104,9 @@ before_script:
   - sleep 10
 
 # command to run tests using coverage, e.g. python setup.py test
-script: tox -e $TOX_ENV -- --elasticsearch
+script:
+  - flake8
+  - tox -e $TOX_ENV -- --elasticsearch
 
 after_success:
   - codecov -e $TOX_ENV
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 
-from django.db import models
 from django.core.paginator import Paginator
+from django.db import models
 from django.utils.six import add_metaclass, iteritems
 from elasticsearch.helpers import bulk
 from elasticsearch_dsl import DocType as DSLDocType
@@ -25,6 +25,7 @@ from .fields import (
 from .indices import Index
 from .registries import registry
 from .search import Search
+from .utils import batch_iterate
 
 model_field_class_to_field_class = {
     models.AutoField: IntegerField,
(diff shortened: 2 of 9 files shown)


safwanrahman:
What about using `iterator()` to optimize it?

sabricot:
https://docs.djangoproject.com/en/2.1/ref/models/querysets/#django.db.models.query.QuerySet.iterator use server-side cursors with postgresql but not with mysql so i guess it wil not optimize every DB backend.
Plus with `iterator` we lose the `prefetch_related()` queries and in some scenarios they can be used.

django_elasticsearch_dsl/documents.py
```diff
@@ -202,13 +202,26 @@ def _prepare_action(self, object_instance, action):
...
-            for page in paginator.page_range:
-                for object_instance in paginator.page(page).object_list:
-                    yield self._prepare_action(object_instance, action)
+        qs_pagination = self._doc_type.queryset_pagination
+        if qs_pagination is not None:
+            if isinstance(object_list, models.QuerySet):
+                # Django paginator can lead to performance issue with huge
+                # queryset, make custom one based on pks presents in db right
+                # now
+                pks = list(object_list.values_list('pk', flat=True))
```
sabricot: I can't really test it. Do you have tested to retrieve millions of pks in this list ?
And don't you think that this `object_list.values_list` queryset should be ordered by pk ?

.travis.yml
```diff
@@ -96,7 +96,9 @@ before_script:
   - sleep 10
 
 # command to run tests using coverage, e.g. python setup.py test
-script: tox -e $TOX_ENV -- --elasticsearch
+script:
+  - flake8
```
sabricot: 👍

tests/test_documents.py
```diff
@@ -11,46 +11,27 @@
                                                  RedeclaredFieldError)
 from tests import ES_MAJOR_VERSION
 
-
-class Car(models.Model):
```
sabricot: 👍

django_elasticsearch_dsl/documents.py
ahivert: Of course ! improve this code with the new commit (--amend)

ahivert:
That should be better now :)

Issue #31, opened 2017-06-26
Handle connection issues


alexislg2:
Hey, I love your package

What about if we set `ELASTICSEARCH_DSL_AUTOSYNC` to True and then the ES server is down for whatever reason? Then all requests made on the indexed docuements will fail.

Any idea how to implement it differently. My current workaround here is to disable ELASTICSEARCH_DSL_AUTOSYNC and set indexing in a crontab job

sabricot:
Hello.

> I love your package

Thank you !

There are several ways to solve this issue, we could add a global settings for disabled the raise of Elasticsearch error or just disabled it in every case (it's already disabled on delete)

But with the 0.2.0 version you can solve this issue by overriding the https://github.com/sabricot/django-elasticsearch-dsl/blob/master/django_elasticsearch_dsl/signals.py#L62

```
def handle_save(self, sender, **kwargs):
    """Handle save.
    Given an individual model instance, update the object in the index.
    """
    instance = kwargs['instance']
    registry.update(instance, raise_on_error=False)
```

ELASTICSEARCH_DSL_SIGNAL_PROCESSOR = 'YouNewSignalProcessor'

PR #109, opened 2018-06-07
Logerrors

btimby:
This PR address #31 by introducing a setting. `ELASTICSEARCH_DSL_AUTOSYNC_LOG_ERRORS`. This setting is `False` by default (current functionality is preserved). However, when enabled, errors are logged (but not raised) by the `BaseSignalProcessor` and it's subclasses.

I want this option to allow myself to have best effort real time indexing without sinking the ship for every indexing error. This option in combination with periodic rebuild of the indexes should provide the trade-off I am looking for.

### DIFF
--- a/django_elasticsearch_dsl/signals.py
+++ b/django_elasticsearch_dsl/signals.py
@@ -55,16 +55,10 @@ class BaseSignalProcessor(object):
         # Do nothing.
 
     def handle_m2m_changed(self, sender, instance, action, **kwargs):
-        try:
-            if action in ('post_add', 'post_remove', 'post_clear'):
-                self.handle_save(sender, instance)
-            elif action in ('pre_remove', 'pre_clear'):
-                self.handle_pre_delete(sender, instance)
-
-        except (ElasticsearchException, ElasticsearchDslException):
-            if not LOG_ERRORS:
-                raise
-            LOGGER.exception('Error during auto-sync %s, continuing', action)
+        if action in ('post_add', 'post_remove', 'post_clear'):
+            self.handle_save(sender, instance)
+        elif action in ('pre_remove', 'pre_clear'):
+            self.handle_pre_delete(sender, instance)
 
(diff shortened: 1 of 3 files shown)


safwanrahman:
This is interesting! can you explain the use case a bit more?

btimby:
Like #31, my case is that I don't want an ElasticSearch problem to affect database transactions. I don't want an exception in the signal handler to report an error or fail to make the change.

IOW, I want ES errors to be ignored (but logged) in the signal handler. Then I have a cron job that rebuilds the index periodically (to resolve these errors).

This is best-effort real-time synchronization with periodic full synchronization.

django_elasticsearch_dsl/signals.py
```diff
@@ -54,22 +66,42 @@ def handle_save(self, sender, instance, **kwargs):
...
     def handle_delete(self, sender, instance, **kwargs):
         """Handle delete.
 
         Given an individual model instance, delete the object from index.
         """
-        registry.delete(instance, raise_on_error=False)
+        try:
+            registry.delete(instance)
+
+        except Exception:
```
safwanrahman: Whats the reason behind capturing all the `Exception`? I think you can catch only the needed exceptions like above.

btimby: The reasoning is to preserve what the method did previously. It passed a flag: `raise_on_error=False`. I removed that (so errors are raised, so I can log them). Then I emulate what used to happen by stuffing `Exception`.

sabricot: I am agree with @safwanrahman, the `raise_on_error` is used in https://elasticsearch-py.readthedocs.io/en/master/helpers.html#elasticsearch.helpers.bulk and by looking to the code of this method i think we only need to catch  https://github.com/elastic/elasticsearch-py/blob/de8413fdbdc6a207401a8cdc8091bf1dda319dab/elasticsearch/helpers/__init__.py#L13

PR #114, opened 2018-06-21
Addition of Production/TLS Elasticsearch connection configuration to README

mahaffey:
wanted to add some clarity here!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,10 +75,10 @@ Here is an example of a more fully filled out ``ELASTICSEARCH_DSL``:
             'hosts': 'https://admin:password@localhost:9200'
             'use_ssl': True,
 
-            # only if using localhost above! ssl certs generally are associated with a full domain eg: example.com
+            # only if using localhost above! ssl certs generally are associated with a full domain eg: example.com or www.example.com
             'ssl_assert_hostname': False,
 
-            # if you are using a self-signed or unrecognized certificate:
+            # if you are using a self-signed or unrecognized certificate authority:
             'ca_certs': <../../certificate-authority.crt>,
 
             # ssl connections with Elasticsearch are more prone to timing out, default is 10 seconds without any retries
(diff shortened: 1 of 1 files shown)


drfrink:
@mahaffey, Thank you for the work you've done on this, I am confused with the ssl certificate file write up piece,

  # if you are using a self-signed or unrecognized certificate authority:
 'ca_certs': <../../certificate-authority.crt>, 
as <../../certificate-authority.crt> throws an invalid syntax error when I put it into 
settings.py, also I'm assuming <../../certificate-authority.crt> is the cert file path?

mahaffey:
@drfrink 

`<../../certificate-authority.crt>` represents the relative file paths to your certificate chain. 

If you are using something like "LetsEncrypt" to generate a self-signed cert, you should follow the docs on where to find those on your system. If you are using an SSL provider like GoDaddy or Comodo, you should be able to find the CA Certs from their documentation as well.

**Do not leave the value to `'ca_certs'` as `<../../certificate-authority.crt>`. That is an invalid path which will lead to errors.**

From the lower level es python wrapper here: https://elasticsearch-py.readthedocs.io/en/master/connection.html
```
ca_certs – optional path to CA bundle. See https://urllib3.readthedocs.io/en/latest/security.html#using-certifi-with-urllib3 for instructions how to get default set
```

Then from the urllib3 (which is a dependency of elasticsearch-py) docs:
```
Custom SSL certificates
Instead of using certifi you can provide your own certificate authority bundle. This is useful for cases where you’ve generated your own certificates or when you’re using a private certificate authority. Just provide the full path to the certificate bundle when creating a PoolManager:

>>> import urllib3
>>> http = urllib3.PoolManager(
...     cert_reqs='CERT_REQUIRED',
...     ca_certs='/path/to/your/certificate_bundle')
When you specify your own certificate bundle only requests that can be verified with that bundle will succeed. It’s recommended to use a separate PoolManager to make requests to URLs that do not need the custom certificate.
```

Issue #115, opened 2018-07-05
Support elasticsearch-dsl > 6.1.0


piedrahitapablo:
I am not sure about this issue but `django-elasticsearch-dsl` does not work with `elasticsearch-dsl` versions higher than 6.1.0, it shows an import error for `DocType` and `DocTypeMeta`, apparently they were removed in newer versions of `elasticsearch-dsl`.

sulami:
See https://github.com/elastic/elasticsearch-dsl-py/pull/930

matali1:
[elastic/elasticsearch-dsl-py#930 ](elastic/elasticsearch-dsl-py#930 ) fixed problem with `DocType` but not `DocTypeMeta` so new version 6.2.1 still doesn't work

safwanrahman:
This package was using an internal undocumented Metaclass of `elasticsearch-dsl`. I am working on a patch to not use the internal undocumented metaclass so its possible to use the latest version of `elasticsearch-dsl`

lthibault:
@safwanrahman Any word on this?  Do you have an ETA?

lthibault:
@safwanrahman No worries!  Happy to hear someone is taking care of it 😃 
Good luck with the major refactor -- we've all been there!

PR #117, opened 2018-07-07
Use Django built-in 'import_string'

dsanders11:
There's a built-in `import_string` in Django since 1.7, so there's no need for the custom `import_class`.


dsanders11:
Build failures are related to #115, not this change.

sabricot:
Thank you and sorry for the delay

PR #121, opened 2018-07-10
Changed setUp and tearDown to classmethods setUpClass and tearDownClass

macher91:
To make more generic creating tests I changed setUp and tearDown to classmethods setUpClass and tearDownClass. Now you don't need create setUp and remeber to put everywhere ex. super().setUp() at the top of setUp()


v1kku:
With this change the index is created once and kept around until python exits. Definitely not the behavior you want. The index should be created at the beginning of each test and deleted after each test. The index should not be reused across tests.

PR #138, opened 2018-11-24
Added http_auth example in README

xaneem:
An example for "http_auth" with username and password was added in README.
"hosts" can accept arrays and the example value was changed to an array.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,8 @@ For example:
 
     ELASTICSEARCH_DSL={
         'default': {
-            'hosts': 'localhost:9200'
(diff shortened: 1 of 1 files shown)


avinashjoshi:
This is exactly what i was looking for! 💯

singhravi1:
Why is this not merged yet?

alaminopu:
@xaneem Probably not in the scope of this project. This is already documented in [`elasticsearch-py`](https://elasticsearch-py.readthedocs.io/en/master/index.html?highlight=http_auth#ssl-and-authentication) client

PR #141, opened 2018-12-15
Add Decimal Field to model to field mapping

PaiAkshay998:
Ran into an issue recently where I had to re-declare a Decimal Field (which I feel is frequently used). I was unsure about the need of creating an entirely different ```Field``` for Decimal so right now, it simply uses Double.


pySilver:
@sabricot  Is there any obvious reason why it haven't been merged?

safwanrahman:
Sorry. I did not notice it. Looks good to me. r+

PR #145, opened 2019-01-05
hook for custom model field class to es dsl field class

pangyun:


safwanrahman:
Merged with #449 including the documentation.

Issue #151, opened 2019-01-25
Only half gets indexed


DannyAziz:
I'm trying to index 800691 `VideoSnippet` documents. When doing `VideoSnippetDocument.search().count()` it returns around `450000` around half.

I've tried again with `queryset_pagination` at about 5000 but it still resulted in the same.

Any help?

DannyAziz:
Changed `queryset_pagination` to 1000 and now I'm getting `504588` for the count

DannyAziz:
Changed `queryset_pagination` to 100 and now I'm getting `506908` for the count. I'm having a feeling that the end count is not tied to `queryset_pagination`

DannyAziz:
Figured it out and created a PR #153.

Without `queryset_pagination` the process was dying as a result of low RAM on my EC2 server. With `queryset_pagination` the qs was not ordered and the pages would not have consistent results.

PR #153, opened 2019-01-28
Queryset pagination order and verbosity

DannyAziz:
Added `.order_by('id')` as was getting inconsistent results after rebuilding the index as shown in #151 

Verbosity isn't important but I liked the feature in 'django-haystack'

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -69,7 +69,8 @@ class DocType(DSLDocument):
         """
         Return the queryset that should be indexed by this doc type.
         """
-        return self.django.model._default_manager.all()
+        primary_key_field_name = self.django.model_meta.pk.name
+        return self.django.model._default_manager.all().order_by(primary_key_field_name)
 
     def prepare(self, instance):
         """
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -206,7 +206,12 @@ class DocType(DSLDocType):
             paginator = Paginator(
                 object_list, self._doc_type.queryset_pagination
             )
+            number_of_pages = len(paginator.page_range)
+            current_page = 0
             for page in paginator.page_range:
+                current_page += 1
+                print("NUMBER OF PAGES: " + str(number_of_pages))
+                print("CURRENT PAGE: " + str(current_page))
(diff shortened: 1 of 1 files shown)


django_elasticsearch_dsl/documents.py
```diff
@@ -206,7 +207,12 @@ def _get_actions(self, object_list, action):
             paginator = Paginator(
                 object_list, self._doc_type.queryset_pagination
             )
+            number_of_pages = len(paginator.page_range)
+            current_page = 0
             for page in paginator.page_range:
+                current_page += 1
+                print("NUMBER OF PAGES: " + str(number_of_pages))
```
crazyscientist: If I may offer some criticism: Maybe you would like to remove the `print` statements before this gets merged?

DannyAziz:
@crazyscientist Are there any other changes that I should make? I would love to see this merged

crazyscientist:
@DannyAziz To me the PR looks good now, but I have no permission to merge/accept.

@sabricot What do you think?

ntcong:
@safwanrahman can you take a look at this? order_by is really important when you have a big dataset. Adding order_by can be a good immediate step before a better approach, like https://github.com/sabricot/django-elasticsearch-dsl/pull/102

safwanrahman:
If the pk is not incremental integer, will this work? I have doubt that in the case the primary key is not integer, like uuid, it will work actually!

safwanrahman:
@DannyAziz Can you rebase it upon master so its possible to merge?

safwanrahman:
Thanks @ntcong for mentioning me here

DannyAziz:
@safwanrahman I will rebase soon 👍 

If you look at these 2 lines:
```
primary_key_field_name = self._doc_type.model._meta.pk.name
return self._doc_type.model._default_manager.all().order_by(primary_key_field_name)
```
It gets the name of the primary key for that model instead of assuming you are using PK, I am using UUID's in my project and it works fine

DannyAziz:
@safwanrahman I've rebased the fork to master would love to see this merged 👍

DannyAziz:
My bad this needs some fixing, hold on...

DannyAziz:
@safwanrahman The tests that are failing don't seem to be related to my changes, any help?

safwanrahman:
This looks good. r+ from my end

Issue #137, opened 2018-11-16
Index build memory and speed issue redux


mjl:
I'm trying to trace down that slow indexing problem that is described in issues #118, #130 and #86.

As a first data point, I have an application that uses django-haystack that I'm trying to move over to DED. Indexing the same miniature data set takes about twice as long and consumes a load of memory.

Here is the indexing performance of haystack with a ~50k test data set:

    $ time python manage.py rebuild_index
    ...
    Indexing 48479 KIT+ plastic

    real	1m16.088s
    user	0m41.056s
    sys	0m1.186s

And here the absolute same for DED:

    $ time python manage.py search_index --rebuild
    ...
    Indexing 48479 'Plastic' objects

    real	1m57.927s
    user	1m8.748s
    sys	0m4.084s

Okay, about half as fast. Not ideal...

Now, on my integration machine, I tried to index a real world dataset of about 1.8M records. I aborted the index run after about 15mins, it used up **all memory** and swap, load was hovering at 18, the machine was totally unresponsive.

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
    666 myself   20   0 3649612 3.330g     16 D   6.6 89.1   2:43.09 python

django-haystack takes about 18m for this same data set. Okay, we really have a problem.

I implemented the same queryset slicing I contributed to haystack (please refer to https://github.com/django-haystack/django-haystack/issues/1077#issuecomment-277677653 for details)
and added some more tweaks to disable Django's queryset caching mechanism.

Again, on the same integration machine as above, here are the results of
different improvements.

[1] Using PagingQuerySet:

This is a proxy queryset that uses iterator() which disables the Django
queryset cache (and uses a server side cursor for fetching data on postgres),
as well as slices the query into multiple parts to work around any caching
in the database driver (this happening is hinted to in the Django docs).

Load stays at around 1, memory usage was virtually constant at a reasonable 60M,
CPU usage is always around 75-80%.

    $ time python manage.py search_index --rebuild -f
    Deleting index 'cc.kitplus-20181116142151'
    Creating index 'cc.kitplus-20181116142151'
    Indexing 1784504 'Plastic' objects

    real	17m38.681s
    user	12m53.976s
    sys	0m14.652s

[2] Using parallel_bulk:

As nothing is actually using the return value from bulk(), I replaced it
with parallel_bulk and returned a dummy value. Errors will be raised as
[...]

PR #154, opened 2019-01-30
Mjl index speedup

mjl:
This pull request addresses the problems when indexing large data sets, both in memory consumption and run time, as discussed in detail in issue #137.

Those changes have been running in a production environment for several months now without any problems.

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 from collections import deque
+from functools import partial
 
 from django.db import models
 from django.utils.six import add_metaclass, iteritems
@@ -200,42 +201,53 @@ class DocType(DSLDocType):
 
     def get_indexing_queryset(self):
         qs = self.get_queryset()
-        # Note: PagingQuerysetProxy handles no chunking, but some tests
-        # check for the qs, so don't interfere.
+        # Note: PagingQuerysetProxy handles the "no chunking" case,
+        #  but some tests check for the mock qs, so don't interfere.
+        #  We could remove this check/branch if the tests are adapted.
         if self._doc_type.queryset_pagination is not None:
             return PagingQuerysetProxy(qs, chunk_size=self._doc_type.queryset_pagination)
         return qs
+
+    def init_prepare(self):
+        """
+        Initialise the data model preparers once here. Extracts the preparers
+        from the model and generate a list of callables to avoid doing that
+        work on every object instance over.
+        """
+        fields = []
+        for name, field in self._doc_type._fields().items():
+            if not isinstance(field, DEDField):
+                continue
+
+            if not field._path:
+                field._path = [name]
+
+            prep_func = getattr(self, 'prepare_%s_with_related' % name, None)
+            if prep_func:
+                fn = partial(prep_func, related_to_ignore=self._related_instance_to_ignore)
+            else:
+                prep_func = getattr(self, 'prepare_%s' % name, None)
+                if prep_func:
+                    fn = prep_func
+                else:
+                    fn = partial(field.get_value_from_instance, field_value_to_ignore=self._related_instance_to_ignore)
+
+            fields.append((name, field, fn))
+
+        self._doc_type._prepared_fields = fields
 
     def prepare(self, instance):
         """
         Take a model instance, and turn it into a dict that can be serialized
         based on the fields defined on this DocType subclass
         """
-        data = {}
-        for name, field in iteritems(self._doc_type._fields()):
-            if not isinstance(field, DEDField):
-                continue
+        if getattr(self._doc_type, '_prepared_fields', None) is None:
+            self.init_prepare()
 
-            if field._path == []:
-                field._path = [name]
-
-            prep_func = getattr(self, 'prepare_%s_with_related' % name, None)
-            if prep_func:
-                field_value = prep_func(
-                    instance,
-                    related_to_ignore=self._related_instance_to_ignore
-                )
-            else:
-                prep_func = getattr(self, 'prepare_%s' % name, None)
-                if prep_func:
-                    field_value = prep_func(instance)
-                else:
-                    field_value = field.get_value_from_instance(
-                        instance, self._related_instance_to_ignore
-                    )
-
-            data[name] = field_value
-
+        data = {
+            name: prep_func(instance)
+                for name, field, prep_func in self._doc_type._prepared_fields
+            }
         return data
 
     @classmethod
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -93,7 +93,6 @@ class PagingQuerysetProxy(object):
             # to values above; this optimises the query for Postgres as not to
             # devolve into multi-second run time at large offsets.
             if self.chunk_size:
-                print("chunk", last_max_pk)
                 if last_max_pk is not None:
                     current_qs = small_cache_qs.filter(pk__gt=last_max_pk)[:self.chunk_size]
                 else:
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -131,6 +131,7 @@ class DocTypeMeta(DSLDocTypeMeta):
         model_field_names = getattr(meta, "fields", [])
         related_models = getattr(meta, "related_models", [])
         queryset_pagination = getattr(meta, "queryset_pagination", None)
+        parallel_indexing = getattr(meta, "parallel_indexing", False)
 
         class_fields = set(
             name for name, field in iteritems(attrs)
@@ -145,6 +146,7 @@ class DocTypeMeta(DSLDocTypeMeta):
         doc_type.auto_refresh = auto_refresh
         doc_type.related_models = related_models
         doc_type.queryset_pagination = queryset_pagination
+        doc_type.parallel_indexing = parallel_indexing
 
         fields = model._meta.get_fields()
         fields_lookup = dict((field.name, field) for field in fields)
@@ -272,6 +274,13 @@ class DocType(DSLDocType):
         deque(parallel_bulk(client=self.connection, actions=actions, **kwargs), maxlen=0)
         return (1, [])  # Fake return value to emulate bulk(), not used upstream
 
+    def _bulk(self, *args, parallel=False, **kwargs):
+        """Helper for switching between normal and parallel bulk operation"""
+        if parallel:
+            return self.parallel_bulk(*args, **kwargs)
+        else:
+            return self.bulk(*args, **kwargs)
+
     def _prepare_action(self, object_instance, action):
         return {
             '_op_type': action,
@@ -301,6 +310,6 @@ class DocType(DSLDocType):
         else:
             object_list = thing
 
-        return self.parallel_bulk(
-            self._get_actions(object_list, action), **kwargs
+        return self._bulk(self._get_actions(object_list, action),
+            parallel=self._doc_type.parallel_indexing, **kwargs
         )
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -1,12 +1,15 @@
 from __future__ import unicode_literals
 
+from collections import deque
+
 from django.db import models
-from django.core.paginator import Paginator
 from django.utils.six import add_metaclass, iteritems
-from elasticsearch.helpers import bulk
+
+from elasticsearch.helpers import bulk, parallel_bulk
 from elasticsearch_dsl import DocType as DSLDocType
 from elasticsearch_dsl.document import DocTypeMeta as DSLDocTypeMeta
 from elasticsearch_dsl.field import Field
+
 
 from .apps import DEDConfig
 from .exceptions import ModelFieldNotMappedError, RedeclaredFieldError
@@ -254,6 +257,10 @@ class DocType(DSLDocType):
     def bulk(self, actions, **kwargs):
         return bulk(client=self.connection, actions=actions, **kwargs)
 
+    def parallel_bulk(self, actions, **kwargs):
+        deque(parallel_bulk(client=self.connection, actions=actions, **kwargs), maxlen=0)
+        return (1, [])  # Fake return value to emulate bulk(), not used upstream
+
     def _prepare_action(self, object_instance, action):
         return {
             '_op_type': action,
@@ -283,6 +290,6 @@ class DocType(DSLDocType):
         else:
             object_list = thing
 
-        return self.bulk(
+        return self.parallel_bulk(
             self._get_actions(object_list, action), **kwargs
         )
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -1,11 +1,13 @@
 from __future__ import unicode_literals
 
+from collections import deque
 from copy import deepcopy
+from functools import partial
 
 from django.db import models
 from django.core.paginator import Paginator
 from django.utils.six import add_metaclass, iteritems
-from elasticsearch.helpers import bulk
+from elasticsearch.helpers import bulk, parallel_bulk
 from elasticsearch_dsl import Document as DSLDocument
 from elasticsearch_dsl.field import Field
 
@@ -50,6 +52,62 @@ model_field_class_to_field_class = {
     models.URLField: TextField,
 }
 
+class PagingQuerysetProxy(object):
+    """
+    I am a tiny standin for Django Querysets that implements enough of
+    the protocol (namely count() and __iter__) to be useful for indexing
+    large data sets.
+
+    When iterated over, I will:
+        - use qs.iterator() to disable result set caching in queryset.
+        - chunk fetching the results so that caching in database driver
+            (especially psycopg2) is kept to a minimum, and database
+            drivers that do not support streaming (eg. mysql) do not
+            need to load the whole dataset at once.
+    """
+    def __init__(self, qs, chunk_size=10000):
+        self.qs = qs
+        self.chunk_size = chunk_size
+
+    def count(self):
+        """Pass through to underlying queryset"""
+        return self.qs.count()
+
+    def __iter__(self):
+        """Iterate over result set. Internally uses iterator() as not
+        to cache in the queryset; also supports chunking fetching data
+        in smaller sets so that databases that do not use server side
+        cursors (django docs say only postgres and oracle do) or other
+        optimisations keep memory consumption manageable."""
+
+        last_max_pk = None
+
+        # Get a clone of the QuerySet so that the cache doesn't bloat up
+        # in memory. Useful when reindexing large amounts of data.
+        small_cache_qs = self.qs.order_by('pk')
+
+        once = no_data = False
+        while not no_data and not once:
+            # If we got the max seen PK from last batch, use it to restrict the qs
+            # to values above; this optimises the query for Postgres as not to
+            # devolve into multi-second run time at large offsets.
+            if self.chunk_size:
+                if last_max_pk is not None:
+                    current_qs = small_cache_qs.filter(pk__gt=last_max_pk)[:self.chunk_size]
+                else:
+                    current_qs = small_cache_qs[:self.chunk_size]
+            else: # Handle "no chunking"
+                current_qs = small_cache_qs
+                once = True	 # force loop exit after fetching all data
+
+            no_data = True
+            for obj in current_qs.iterator():
+                # Remember maximum PK seen so far
+                last_max_pk = obj.pk
+                no_data = False
+                yield obj
+
+            current_qs = None  # I'm free!
 
 class DocType(DSLDocument):
     def __init__(self, related_instance_to_ignore=None, **kwargs):
@@ -77,36 +135,55 @@ class DocType(DSLDocument):
         """
         return self.django.model._default_manager.all()
 
+    def get_indexing_queryset(self):
+        qs = self.get_queryset()
+        # Note: PagingQuerysetProxy handles no chunking, but some tests
+        # check for the qs, so don't interfere.
+        if self.django.queryset_pagination is not None:
+            return PagingQuerysetProxy(qs, chunk_size=self.django.queryset_pagination)
+        return qs
+
+    def init_prepare(self):
+        """
+        Initialise the data model preparers once here. Extracts the preparers
+        from the model and generate a list of callables to avoid doing that
+        work on every object instance over.
+        """
+        fields = []
+        for name, field in iteritems(self._fields):
+            if not isinstance(field, DEDField):
+                continue
+
+            if not field._path:
+                field._path = [name]
+
+            prep_func = getattr(self, 'prepare_%s_with_related' % name, None)
+            if prep_func:
+                fn = partial(prep_func, related_to_ignore=self._related_instance_to_ignore)
+            else:
+                prep_func = getattr(self, 'prepare_%s' % name, None)
+                if prep_func:
+                    fn = prep_func
+                else:
+                    fn = partial(field.get_value_from_instance, field_value_to_ignore=self._related_instance_to_ignore)
+
+            fields.append((name, field, fn))
+
+        self._doc_type._prepared_fields = fields
+
     def prepare(self, instance):
         """
         Take a model instance, and turn it into a dict that can be serialized
         based on the fields defined on this DocType subclass
         """
-        data = {}
-        for name, field in iteritems(self._fields):
-            if not isinstance(field, DEDField):
-                continue
+        if getattr(self._doc_type, '_prepared_fields', None) is None:
+            self.init_prepare()
 
-            if field._path == []:
-                field._path = [name]
-
-            prep_func = getattr(self, 'prepare_%s_with_related' % name, None)
-            if prep_func:
-                field_value = prep_func(
-                    instance,
-                    related_to_ignore=self._related_instance_to_ignore
-                )
-            else:
-                prep_func = getattr(self, 'prepare_%s' % name, None)
-                if prep_func:
-                    field_value = prep_func(instance)
-                else:
-                    field_value = field.get_value_from_instance(
-                        instance, self._related_instance_to_ignore
-                    )
-
-            data[name] = field_value
-
+        data = {
+            name: prep_func(instance)
+                for name, field, prep_func in self._doc_type._prepared_fields
+            }
+        # print("-> %s" % data)
         return data
 
     @classmethod
@@ -128,6 +205,10 @@ class DocType(DSLDocument):
     def bulk(self, actions, **kwargs):
         return bulk(client=self._get_connection(), actions=actions, **kwargs)
 
+    def parallel_bulk(self, actions, **kwargs):
+        deque(parallel_bulk(client=self._get_connection(), actions=actions, **kwargs), maxlen=0)
+        return (1, [])  # Fake return value to emulate bulk(), not used upstream
+
     def _prepare_action(self, object_instance, action):
         return {
             '_op_type': action,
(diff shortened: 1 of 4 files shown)


barseghyanartur:
@sabricot:

+1 for merge!

@mjl:

Thank you!

pySilver:
any updates on this?

barseghyanartur:
Come on, guys, this is one of the best improvements out there! It's a shame it's not merged yet.

safwanrahman:
Sorry for taking so long. @mjl Can you remove the merge commit and rebase your changes upon master?

django_elasticsearch_dsl/documents.py
```diff
@@ -44,6 +48,62 @@
...
+        small_cache_qs = self.qs.order_by('pk')
+
+        once = no_data = False
+        while not no_data and not once:
+            # If we got the max seen PK from last batch, use it to restrict the qs
+            # to values above; this optimises the query for Postgres as not to
+            # devolve into multi-second run time at large offsets.
+            if self.chunk_size:
+                if last_max_pk is not None:
+                    current_qs = small_cache_qs.filter(pk__gt=last_max_pk)[:self.chunk_size]
```
safwanrahman: What about when the `primary key` is not integer?

safwanrahman:
I have gone through the code and it seems like it needs some improvement.

@barseghyanartur I understand your frustration, but this thing works for you does not mean will work for everyone. While maintaining packages, you need to make it generalized. So its not so easy to merge anything!

barseghyanartur:
@safwanrahman:

I haven't seen breaking changes on the API level here. Everything stays on as it was.

safwanrahman:
Its not breaking chnages in API level, but you need to ensure that it works properly for all the people who are using it. So its not a easy call

mjl:
> What about when the `primary key` is not integer?

   It doesn't matter what type the primary key is, as long as it's 
sortable by the database. Integer, Strings, guuid, all should work fine.

   Also, that feature is opt-in, so if one really runs into problems, 
just don't enable it?

safwanrahman:
@mjl You always can not ensure with `uuid` that the later one is always greater as its not incremental. Moreover, if its unique string, you always can not ensure that the later one will be greater. As you are making `gte` query, the later one will not be indexed.

mjl:
> @mjl You always can not ensure with `uuid` that the later one is 
> always greater as its not incremental. Moreover, if its unique string, 
> you always can not ensure that the later one will be greater. As you 
> are making `gte` query, the later one will not be indexed.

   That is only a problem if you are adding data while doing a full 
reindex and expect the full reindex to include those entries. That, to 
be blunt, I would call user error, as iterating over changing data sets 
is always brittle. And note that won't even work as you expect with the 
current code: Anything that is added while the index is rebuilt will not 
be included in the index as the sql query is done at a certain point in 
time (and thus its result set frozen); anything added after that time 
will not be indexed by the rebuild command. You might catch that case by 
using signals or whatnot, but that is outside the scope of the indexing 
code.

   But fair enough, perhaps let's add some warning in the docs? What 
should it say?

safwanrahman:
> That is only a problem if you are adding data while doing a full
reindex and expect the full reindex to include those entries. That, to
be blunt, I would call user error, as iterating over changing data sets
is always brittle. And note that won't even work as you expect with the
current code: Anything that is added while the index is rebuilt will not
be included in the index as the sql query is done at a certain point in
time (and thus its result set frozen); anything added after that time
will not be indexed by the rebuild command. You might catch that case by
using signals or whatnot, but that is outside the scope of the indexing
code.

   But fair enough, perhaps let's add some warning in the docs? What
should it say?

@mjl I understand your point. I think adding documentation will be better idea.

I have a feeling to not change the `get_queryset` method. As its a public method, we should not change it.
According to [documentation](https://docs.djangoproject.com/en/2.2/ref/models/querysets/), you could use `iterator` with `chunk_size`. Django would manage everything that you are going to do actually. Either it will use the database level cursor, or load it in memory.

django_elasticsearch_dsl/documents.py
```diff
@@ -44,6 +48,62 @@
...
+                else:
+                    current_qs = small_cache_qs[:self.chunk_size]
```
safwanrahman: Whats the reason to slice before running the `iterator`? [Django handle it by default if you pass certain ](https://docs.djangoproject.com/en/2.2/ref/models/querysets/) chunk_size`

django_elasticsearch_dsl/documents.py
```diff
@@ -134,16 +217,16 @@ def _prepare_action(self, object_instance, action):
...
-            for page in paginator.page_range:
-                for object_instance in paginator.page(page).object_list:
-                    yield self._prepare_action(object_instance, action)
+        for object_instance in object_list:
+            yield self._prepare_action(object_instance, action)
+
+    def _bulk(self, *args, **kwargs):
+        """Helper for switching between normal and parallel bulk operation"""
+        parallel = kwargs.pop('parallel', False)
+        if parallel:
```
safwanrahman: We should find a way where while updating a single object the `parallel` is always false. We need to use it only while indexing large number of objects. When a model object is updated, the django signal call the same method for updating the index.

mjl:
> +            # If we got the max seen PK from last batch, use it to 
> restrict the qs
> +            # to values above; this optimises the query for Postgres 
> as not to
> +            # devolve into multi-second run time at large offsets.
> +            if self.chunk_size:
> +                if last_max_pk is not None:
> +                    current_qs = 
> small_cache_qs.filter(pk__gt=last_max_pk)[:self.chunk_size]
> +                else:
> +                    current_qs = small_cache_qs[:self.chunk_size]


> Whats the reason to slice before running the `iterator`? [Django 
> handle it by default if you pass certain 
> ](https://docs.djangoproject.com/en/2.2/ref/models/querysets/) 
> chunk_size`

   Somewhere in the interaction between django and postgres, iterating 
over large datasets produces very (VERY) long query times. In my tests, 
the first couple of batches came in at 200ms per query, and later 
batches came in at 12000ms. That problem might or might not have been 
fixed in the meantime, I really do not know. The code has originated in 
django-haystack in 2017, so things might be better now (one can always 
hope!).

   I have analyzed the problem in 
https://github.com/django-haystack/django-haystack/issues/1077#issuecomment-277677653

mjl:
> We should find a way where while updating a single object the 
> `parallel` is always false. We need to use it only while indexing 
> large number of objects. When a model object is updated, the django 
> signal call the same method for updating the index.

   Isn't the _bulk thingy only used by the reindex command?

safwanrahman:
> Isn't the _bulk thingy only used by the reindex command?

unfortunately no! I also wish the `_bulk` would be called in reindex only, but it seems like while indexing a single object, `_bulk` is called. 😢

safwanrahman:
> Somewhere in the interaction between django and postgres, iterating
over large datasets produces very (VERY) long query times. In my tests,
the first couple of batches came in at 200ms per query, and later
batches came in at 12000ms. That problem might or might not have been
fixed in the meantime, I really do not know. The code has originated in
django-haystack in 2017, so things might be better now (one can always
hope!).

I think the problem might have fixed in the meantime. I have use `iterator` in my workplace for querying a table hacing millions of data, but did not face any problem. Let me know if the issue still persist so I can investigate more.

mjl:
> I think the problem might have fixed in the meantime. I have use 
> `iterator` in my workplace for querying a table hacing millions of 
> data, but did not face any problem.

   Ok, you mean just remove the slicing action and use the iterator 
directly? I can try that and compare it, no problem.

mjl:
> unfortunately no! I also wish the `_bulk` would be called in reindex 
> only, but it seems like while indexing a single object, `_bulk` is 
> called. 😢

   Ok, let's see if we can distinguish those cases...

safwanrahman:
> Ok, you mean just remove the slicing action and use the iterator
directly? I can try that and compare it, no problem.

Yes. I think we can remove the `PagingQuerysetProxy` object and handle it directly in the `Document`. what do you think?

mjl:
> Yes. I think we can remove the `PagingQuerysetProxy` object and handle 
> it directly in the `Document`. what do you think?

   I'll try and report back. I'm all for removing unnecessary 
complexity!

pySilver:
@mjl sry to bother you :) any news from your side?

mjl:
> @mjl sry to bother you :) any news from your side?

   Yes, thanks for the poke. I've done some measurements that I need to 
bring into understandable format, but in a nutshell, it really looks as 
if the paging part is no longer necessary. I'll summarize what I have in 
a bit.

pySilver:
@mjl actually there is some issue when using non standard queryset:

```python
DISTINCT_BY_FIELDS = [
        "item_group_id",
        "brand_id",
        "merchant_id",
        "material",
        "pattern",
        "color",
        "adult",
        "multipack",
        "is_bundle",
        "age_group",
        "gender",
        "size_type",
        "size_system",
    ]
    def get_queryset(self):
        # In case of performance issues we should use `CROSS  JOIN LATERAL`
        # @see: https://medium.com/kkempin/postgresqls-lateral-join-bfd6bd0199df
        # @see: https://heap.io/blog/engineering/postgresqls-powerful-new-join-type-lateral
        # @see: https://dbfiddle.uk/?rdbms=postgres_11&fiddle=68cc26893cb96a912997ac6342f42490
        return (
            super()
            .get_queryset()
            .select_related("brand", "merchant", "coupon")
            .distinct(*self.DISTINCT_BY_FIELDS)
            .order_by(*self.DISTINCT_BY_FIELDS)
        )
```

```Are you sure you want to delete the '<elasticsearch_dsl.index.Index object at 0x10c481190>' indexes? [n/Y]: y
Deleting index '<elasticsearch_dsl.index.Index object at 0x10c481190>'
INFO 2019-10-01 16:33:42,030 base 90395 4591814080 DELETE http://localhost:9200/products [status:200 request:0.039s]
Creating index '<elasticsearch_dsl.index.Index object at 0x10c481190>'
INFO 2019-10-01 16:33:42,108 base 90395 4591814080 PUT http://localhost:9200/products [status:200 request:0.075s]
Indexing 1003 'Product' objects
Traceback (most recent call last):
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.InvalidColumnReference: SELECT DISTINCT ON expressions must match initial ORDER BY expressions
LINE 1: ...O SCROLL CURSOR WITH HOLD FOR SELECT DISTINCT ON ("catalog_p...
                                                             ^


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1100, in execute_sql
    cursor.execute(sql, params)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/backends/utils.py", line 99, in execute
    return super().execute(sql, params)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/backends/utils.py", line 76, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.ProgrammingError: SELECT DISTINCT ON expressions must match initial ORDER BY expressions
LINE 1: ...O SCROLL CURSOR WITH HOLD FOR SELECT DISTINCT ON ("catalog_p...
                                                             ^


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./manage.py", line 30, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 134, in handle
    self._rebuild(models, options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 115, in _rebuild
    self._populate(models, options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 92, in _populate
    doc().update(qs)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/documents.py", line 248, in update
    **kwargs
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/documents.py", line 227, in _bulk
    return self.parallel_bulk(*args, **kwargs)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/documents.py", line 206, in parallel_bulk
    deque(parallel_bulk(client=self._get_connection(), actions=actions, **kwargs), maxlen=0)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/elasticsearch/helpers/actions.py", line 365, in parallel_bulk
    actions, chunk_size, max_chunk_bytes, client.transport.serializer
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 748, in next
    raise value
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 140, in _helper_reraises_exception
    raise ex
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 292, in _guarded_task_generation
    for i, x in enumerate(iterable):
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/elasticsearch/helpers/actions.py", line 63, in _chunk_actions
    for action, data in actions:
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/documents.py", line 220, in _get_actions
    for object_instance in object_list:
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/documents.py", line 100, in __iter__
    for obj in current_qs.iterator():
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/models/query.py", line 341, in _iterator
    yield from self._iterable_class(self, chunked_fetch=use_chunked_fetch, chunk_size=chunk_size)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/models/query.py", line 55, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1103, in execute_sql
    cursor.close()
psycopg2.errors.InvalidCursorName: cursor "_django_curs_123145595777024_1" does not exist```

**Comment by mjl** [comment]
Ah I see. As the PagingProxy tacks its own order_by() onto the 
queryset, that of course does no longer match the distinct() you add in 
get_queryset(), and thus it blows up with

> psycopg2.errors.InvalidColumnReference: SELECT DISTINCT ON expressions 
> must match initial ORDER BY expressions

   On the positive side, all that manual slicing seems unnecessary in 
recent djangos/psycopg2. Thanks for the report -- may I ping you for a 
new version to try later?

   In the meantime, you can try to do the following: In 
documents.py:135, make get_indexing_queryset() just return 
self.get_queryset().iterator(). That's untested, but should disable the 
PagingQuerysetProxy altogether and be pretty much what I'm thinking 
should work just as well.

**Comment by pySilver** [comment]
@mjl sure, thanks for heads up! if you really need specific ordering you can append a column to existing ordering fields (postgresql requires prefix of distinct and orderby to be the same)

**Comment by pySilver** [comment]
Returning queryset.iterator() is not working due to:

```
INFO 2019-10-01 17:08:15,979 base 90722 4596716992 PUT http://localhost:9200/products [status:200 request:0.084s]
Traceback (most recent call last):
  File "./manage.py", line 30, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 134, in handle
    self._rebuild(models, options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 115, in _rebuild
    self._populate(models, options)
  File "/Users/Silver/.local/share/virtualenvs/mybaze-dsw6HN8p/lib/python3.7/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 90, in _populate
    qs.count(), doc.django.model.__name__)
AttributeError: 'generator' object has no attribute 'count'
```

**Comment by pySilver** [comment]
Plain `return self.get_queryset()` works.

**Comment by pySilver** [comment]
In order to make it work with `PagingQuerysetProxy`  the following should be changed:

at `django_elasticsearch_dsl.documents.PagingQuerysetProxy.__iter__`

from:
`small_cache_qs = self.qs.order_by('pk')`
to:
`small_cache_qs = self.qs.order_by(self.qs.query.order_by + ('pk',))`
(so it preserves ordering provided in orignal qs if any)

**Comment by mjl** [comment]
No, using a different ordering will not work correctly. And not using iterator will consume unbelievable amounts of memory, so you miss out of most of the benefits.

**Comment by pySilver** [comment]
hm why different ordering won't work?

**Comment by mjl** [comment]
Because the PagingQuerysetProxy chops up the result set ordered by pk. If your queryset is not ordered by pk, then a full reindex will certainly not index the whole dataset. That's the reason the order_by('pk') that caused your exception in the first place is there.

I'll keep you updated when I have something new.

**Comment by pySilver** [comment]
im not 100% sure but as I understand `distinct on` in my case it should not have any difference. results are sorted by pk, preceding ordering keys are needed for db to eliminate duplicates. but again, I cannot test it.

**Comment by mjl** [comment]
Okay, I have a small analysis of the state of affairs here...

### Plain ded 6.4.2

    Deleting index 'kit-20191002100942'
    Creating index 'kit-20191002100942'
    Indexing 406304 'Plastic' objects

    real	31m12.357s
    user	5m59.528s
    sys	0m10.205s

### Plain ded 6.4.2 plus .iterator() in _populate

```
Deleting index 'kit-20191002104856'
Creating index 'kit-20191002104856'
Indexing 406304 'Plastic' objects

real	12m25.691s
user	5m56.511s
sys	0m8.887s
```

### ded mjl-index-speedup-6.4.2

```
Deleting index 'kit-20191002110406'
Creating index 'kit-20191002110406'
Indexing 406304 'Plastic' objects

real	6m56.811s
user	4m27.797s
sys	0m7.669s
```

### ded mjl-index-speedup-6.4.2 with PagingQuerysetProxy ripped out

```
Deleting index 'kit-20191002114453'
Creating index 'kit-20191002114453'
Indexing 406304 'Plastic' objects

real	6m4.546s
user	4m27.598s
sys	0m9.618s
```

### Same, but remove order_by from model.Meta

```
Deleting index 'kit-20191002122409'
Creating index 'kit-20191002122409'
Indexing 406304 'Plastic' objects

real	5m26.150s
user	4m26.708s
sys	0m8.720s
```

### Memory profiling

Figure_1

Also interesting is the memory consumption: as you can see, iterator() reduces memory pressure considerably. And also, you see a flat "step" in most of the graphs, that is the database sorting the result set, thus my last test without an model.Meta.ordering.

@pySilver, can you try the latest from the mjl-index-speedup-6 branch? That should be at ccf4666 right now? I have not updated the -7 branch yet, because I don't have an ES7 to test with.

**Comment by pySilver** [comment]
@mjl wow, thats a nice analysis you've made!

Unfortunately I cannot test on 6x branch since we are using 7x :/ I'll take a look at changes and apply to 7.x

**Comment by mjl** [comment]
> Unfortunately I cannot test on 6x branch since we are using 7x :/ I'll 
> take a look at changes and apply to 7.x

   Ok, that was easy, try mjl-index-speedup-7. It's a bit behind the 
official ded-7, but might be enough to get you going?

**Comment by pySilver** [comment]
Yes, I'll handle this in a few mins. Is there anything you'd like me to check particularly?

**Comment by pySilver** [comment]
if there a bench script – I can run it

**Comment by mjl** [comment]
No, just try whatever you feel you need to try. You seem to have pretty involved model structure and queries, so if it works for you, it probably works for pretty much everybody else :-)

Let me know how it goes, and if it helps, speedwise. Thanks!

**Comment by pySilver** [comment]
@mjl it work perfectly, thank you!

@sabricot isn't it ready to be merged into master?

**Comment by safwanrahman** [comment]
@pySilver I have given some fixup for this PR, once its fixed, it will be able to merge! :D

**Comment by safwanrahman** [comment]
@mjl do you need any kind of support to fix the issues I have mentioned? Feel free to let me know.

**Comment by pySilver** [comment]
@mjl @safwanrahman  

I've checked current code and 2 out of 3 issues mentioned by @safwanrahman looks outdated at the moment. The one left is about `django_elasticsearch_dsl.documents.DocType._bulk` which apparently is called for single item's updates (signals). Is it still relevant? Maybe simply propagate some flag from management command there?

**Comment by mjl** [comment]
> The one left is about 
> `django_elasticsearch_dsl.documents.DocType._bulk`

That's a coincidence, I just looked into that... _bulk() only switches 
to parallel mode if a parallel=True kwarg is present. This comes through 
update() and is copied there from the document's options. In order to 
make that only set for a full reindex, I'd have to pass that parameter 
down from management/commands/search_index _populate(), which sounds not 
too bad.

Basically, three smallish changes. Would that do the trick @sabricot?

```
@@ -172,7 +172,7 @@ class DocType(DSLDocument):
          else:
              return self.bulk(*args, **kwargs)

-    def update(self, thing, refresh=None, action='index', **kwargs):
+    def update(self, thing, refresh=None, action='index', 
may_parallelise=False, **kwargs):
          """
          Update each document in ES for a model, iterable of models or 
queryset
          """
@@ -188,7 +188,7 @@ class DocType(DSLDocument):

          return self._bulk(
              self._get_actions(object_list, action),
-            parallel=self.django.parallel_indexing,
+            parallel=may_parallelise and self.django.parallel_indexing,
              **kwargs
          )

diff --git 
a/django_elasticsearch_dsl/management/commands/search_index.py 
b/django_elasticsearch_dsl/management/commands/search_index.py
index e4cc7ef..00d25a9 100644
--- a/django_elasticsearch_dsl/management/commands/search_index.py
+++ b/django_elasticsearch_dsl/management/commands/search_index.py
@@ -89,7 +89,7 @@ class Command(BaseCommand):
                  doc().get_queryset().count(), 
doc.django.model.__name__)
              )
              qs = doc().get_indexing_queryset()
-            doc().update(qs)
+            doc().update(qs, may_parallelise=True)

```

safwanrahman:
@mjl I think it can be done. But I have mentioned couple of issues above that need to be fixed.
@pySilver There are some more issue to fix like removing the slicing and remove the extreme complexity.

safwanrahman:
I think its better to pass a argument while reindexing rathar than setting parallel index in document level.

mjl:
> @mjl I think it can be done. But I have mentioned couple of issues 
> above that need to be fixed.

   I went over this thread and all I think all the issues you mentioned 
have been resolved (by ripping out the PagingQuerysetProxy). Did I miss 
some?

mjl:
> I think its better to pass a argument while reindexing rathar than 
> setting parallel index in document level.

But then you would have to set that argument every time you do a 
reindex? I think a parameter makes sense in addition (so you can "try it 
out" without having to change things), but it's kinda nice you can set a 
default in the document...?

safwanrahman:
@mjl Its only used the time when anyone run `populate` or `reindex` management command. I think, it does not make sense to have it in the document level because its not related to document. If we add the configuration in document level, people may think that its used while indexing single object also, but according to our approach, its not the case.

safwanrahman:
@mjl Following work is needed actually, then I can give another review.
- Removing `PagingQuerysetProxy`
- Fixing the queryset slicing
- management command argument for paraller indexing

mjl:
Okay, I see you point. Still I like the "set a default and forget" 
convenience; so perhaps rename it to parallel_only_used_on_full_reindex 
:-) or somesuch?

mjl:
> - Removing `PagingQuerysetProxy`
> - Fixing the queryset slicing

Those are done.

> - management command argument for paraller indexing

Give me a minute :-)

safwanrahman:
> Okay, I see you point. Still I like the "set a default and forget"
convenience; so perhaps rename it to parallel_only_used_on_full_reindex
:-) or somesuch?

What about adding a value in settings where people can add the default one? I think adding a settings makes more sense if you want something like add one time and forget! 😉

mjl:
Fair enough! I've added ELASTICSEARCH_DSL_PARALLEL to settings and 
--parallel to the search_index management command.

Let me know if anything else needs adjusting!

safwanrahman:
@mjl Have you pushed your changes? I dont see it updated!

safwanrahman:
I think you have pushed to other branch, not this branch!

mjl:
Ah sorry, I'll incorporate the changes in this branch in a sec.

pySilver:
I agree with @sabricot that parallel does not belong on document level but management command. And I think it should be optional  flag to follow what ES does (it does not imply bulk indexing by default). Having a config flag for "set and forget" sounds as a good idea.

safwanrahman:
@pySilver @safwanrahman and @sabricot are different person! 😅

mjl:
I had to delete and re-create my branch because it would not rebase easily;
a new pull request was necessary -> #213

PR #160, opened 2019-03-14
Class-based view for Django

crazyscientist:
For one of my projects I wanted to use the search results from elasticsearch in an analogous way like query results. Therefore I have written a view class that can be used to easily provide a view for listing results, e.g.:

```python
class MyView(FormMixin, ElasticListView):
    model = Issue
    document = IssueDocument
    paginate_by = 25
    form_class = SearchForm

    def _get_queryset(self):
        # Build your ES query here
       ...

    ...

    def post(self, request, *args, **kwargs):
        return self.get(request, *args, **kwargs)
```
This is just a demo to give you the idea of the usage.

If you like the idea, I will put some effort in writing documentation and tests for this :)

### DIFF
--- a/django_elasticsearch_dsl/paginator.py
+++ b/django_elasticsearch_dsl/paginator.py
@@ -1,0 +1,38 @@
+from math import ceil
+
+from django.core.paginator import Page, Paginator
+from django.utils.functional import cached_property
+
+
+class ElasticPage(Page):
+    def __len__(self):
+        if hasattr(self.object_list, "count"):
+            return self.object_list.count()
+
+        return len(self.object_list)
+
+
+class ElasticPaginator(Paginator):
+    max_result_window = 10000
+
+    @cached_property
+    def num_pages(self):
+        """
+        Returns the total number of pages.
+
+        We are aware of the ElasticSearch's internal limit for retrieving result
+        pages. Therefore we do not offer more results!
+        """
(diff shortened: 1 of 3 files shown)


abdelwahebmoalla:
@crazyscientist I like the idea please provide more examples

safwanrahman:
Can you add some documentations?

safwanrahman:
This looks really interesting. I will work on this to make it compatable with new version. @crazyscientist I am sorry for the delay, are you still interested to work on this?

crazyscientist:
I am also sorry :laughing: Somewhere between 2019 and now we discovered that a setup with Elasticsearch did not fulfill the expectation of users to search data in real-time. So we do the search in the database again; with some strong query optimization.

safwanrahman:
@crazyscientist Cool, I can understand. Can you share some thoughts about the database search? Which database, how do you optimize, what is the response time?

crazyscientist:
@safwanrahman In our app we now also use Datatables, which is fed by an API based on DRF. Our optimizations for the two most common use cases are implemented in a subclass of `rest_framework_datatables.django_filters.backends.DatatablesFilterBackend`. The idea was basically, to identify the use case based on the search term with a few quick queries and then construct a query with subqueries for that particular use case.
With a Postgres DB a user usually gets results in less than 4 seconds; I know, it's not great, but nobody complains :joy:

PR #168, opened 2019-04-30
Filter an existing django queryset

andriilahuta:
Slight modification of `to_queryset` method to allow filtering of existing django querysets.

### DIFF
--- a/django_elasticsearch_dsl/search.py
+++ b/django_elasticsearch_dsl/search.py
@@ -42,9 +42,16 @@ class Search(DSLSearch):
 
         return queryset
 
+    def _get_queryset(self):
+        """
+        Return a django queryset that will be filtered by to_queryset method.
+        """
+        return self._model._default_manager.all()
+
     def to_queryset(self, keep_order=True):
         """
         Return a django queryset from the elasticsearch result.
         It costs a query to the sql db.
         """
-        return self.filter_queryset(self._model._default_manager, keep_order)
+        qs = self._get_queryset()
+        return self.filter_queryset(qs, keep_order)
--- a/django_elasticsearch_dsl/search.py
+++ b/django_elasticsearch_dsl/search.py
@@ -1,4 +1,5 @@
 from django.db.models import Case, When
+from django.shortcuts import _get_queryset
 
 from elasticsearch_dsl import Search as DSLSearch
 
@@ -13,12 +14,20 @@ class Search(DSLSearch):
         s._model = self._model
         return s
 
-    def to_queryset(self, keep_order=True):
+    def filter_queryset(self, klass, keep_search_order=True):
         """
-        This method return a django queryset from the an elasticsearch result.
-        It cost a query to the sql db.
+        Filter an existing django queryset using the elasticsearch result.
+        It costs a query to the sql db.
+        klass may be a Model, Manager, or QuerySet object.
         """
+
+        qs = _get_queryset(klass)
         s = self
+        if s._model is not qs.model:
+            raise TypeError(
+                'Unexpected queryset model '
+                '(should be: %s, got: %s)' % (s._model, qs.model)
+            )
 
         # Do not query again if the es result is already cached
         if not hasattr(self, '_response'):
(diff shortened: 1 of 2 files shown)


safwanrahman:
Can you please rebase it upon master and also describe the use case of this?

laevilgenius:
@safwanrahman it can be used when implementing a custom `QuerySet` class or searching otherwise pre-filtered queryset.

django_elasticsearch_dsl/search.py
```diff
@@ -13,12 +14,20 @@ def _clone(self):
...
+    def filter_queryset(self, klass, keep_search_order=True):
         """
-        This method return a django queryset from the an elasticsearch result.
-        It cost a query to the sql db.
+        Filter an existing django queryset using the elasticsearch result.
+        It costs a query to the sql db.
+        klass may be a Model, Manager, or QuerySet object.
         """
+
+        qs = _get_queryset(klass)
```
safwanrahman: I am not comfortable to use a private method `_get_queryset` as it is not documented anywhere.

I think it is better to use `self._model._default_manager` for getting the default manager as the [documentation](https://docs.djangoproject.com/en/3.0/topics/db/managers/#default-managers) recommend it.

safwanrahman:
@laevilgenius Sorry for the too late reply. I have one issue with the PR. Are you able to fix it?

django_elasticsearch_dsl/search.py
laevilgenius: Done.

django_elasticsearch_dsl/search.py
```diff
@@ -13,12 +13,17 @@ def _clone(self):
         s._model = self._model
         return s
 
-    def to_queryset(self, keep_order=True):
+    def filter_queryset(self, queryset, keep_search_order=True):
```
safwanrahman: This should be a internal function so it should start with `_`

safwanrahman:
I think there should be a `get_queryset` method like django rest framework, which will be called by the `to_queryset` method. Sot the filtering and others can be done by overwriting `get_queryset` method. 

Can you add a documentation about that also?

django_elasticsearch_dsl/search.py
laevilgenius: I actually meant it to be public, so I could do something like this:
```py
class SearchQuerySet(QuerySet):
    def from_elastic(self, search):
        return search.filter_queryset(self)
```

Could you explain your reasoning?

safwanrahman, line 16: Thanks for the update. As we already have `get_queryset` method, you can make the filter there. So we do not need another method for `filter_queryset`

django_elasticsearch_dsl/search.py
```diff
@@ -27,13 +32,26 @@ def to_queryset(self, keep_order=True):
...
+        if keep_search_order:
             preserved_order = Case(
                 *[When(pk=pk, then=pos) for pos, pk in enumerate(pks)]
             )
-            qs = qs.order_by(preserved_order)
+            queryset = queryset.order_by(preserved_order)
+
+        return queryset
+
+    def _get_queryset(self):
```
safwanrahman, line 45: You can make this api public!

django_elasticsearch_dsl/search.py
safwanrahman: Interesting. I did not think about it before. Let me think about it so I can suggest a better approach.

safwanrahman: I think `to_queryset` should accept a `queryset` as optional parameter. So if `queryset` is passed, it will use that queryset. otherwise call the `get_queryset` method to get the `queryset`

@laevilgenius what do you think?

laevilgenius: Should we keep `to_queryset` backwards compatible? I'm thinking `def to_queryset(self, qs=None, keep_order=True)` vs `def to_queryset(self, keep_order=True, qs=None)`.

django_elasticsearch_dsl/search.py
laevilgenius: I made it private so it would be less confusing with IDE hinting for example, as we have public methods that are mostly related to filtering and search object manipulation. What would be the public usage of `get_queryset`?

safwanrahman:
I think we can merge this PR as it is now. Sorry for the late in reviewing, better late than never! Thanks @andriilahuta

PR #169, opened 2019-04-30
Add elasticsearch online/offline checker

laevilgenius:
Added a helper function to check whether elasticsearch is online. 
This allows to skip integration tests when elasticsearch is down.


safwanrahman:
Thanks @laevilgenius ! Better late than never! 😅

PR #179, opened 2019-06-06
Add requirement to define Document class in documents.py in app dir

Acetonen:
Hi, your package is great! But I think that my one string readme fix is essential for new users. I've spent a few hours to realize this requirement.))
Cheers!


safwanrahman:
Cool! Thanks a lot!

PR #181, opened 2019-06-10
fix README

Acetonen:
Aloha!
It's me again) 
Fix importing DocType from django_elasticsearch_dsl in README.rst


Acetonen:
wat???

safwanrahman:
Thanks for the PR. But I have fixed it in latest version! Thanks!

PR #185, opened 2019-06-15
Import Document not DocType

ogiogi93:
~I think `from django_elasticsearch_dsl import Document` is not a correct path  🤔~

just import Document not DocType

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,7 +84,7 @@ It required to defined ``Document`` class in  ``documents.py`` in your app direc
 
     # documents.py
 
-    from django_elasticsearch_dsl import Document
+    from django_elasticsearch_dsl.documents import Document
     from django_elasticsearch_dsl.registries import registry
     from .models import Car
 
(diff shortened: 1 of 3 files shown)


safwanrahman:
LGTM. r+

Acetonen:
@safwanrahman you already broke my production with this))) 
image

i install django_elasticsearch_dsl directly from git(  because pip version don't update

safwanrahman:
@Acetonen Sorry. How it broke your production?

Acetonen:
@safwanrahman like this:

image

I've fixed it, but I think you have to leave here:

image

this: 
`from .documents import Document, DocType`
until you get stable and updated PIP version

safwanrahman:
@Acetonen You must install a specific version as mentioned in the doc not the master branch.
`pip install https://github.com/sabricot/django-elasticsearch-dsl/archive/6.4.1.tar.gz`

Acetonen:
@safwanrahman yea, my fault(

PR #206, opened 2019-09-26
Fix name resolution

pySilver:
elasticsearch_dsl.index.Index cannot be casted to string, `_name` property should be used instead.

Otherwise you'll see this in console:
```Are you sure you want to delete the '<elasticsearch_dsl.index.Index object at 0x106d95e50>' indexes? [n/Y]: y```

Expected:
```Are you sure you want to delete the 'index_name_here' indexes? [n/Y]: y```


safwanrahman:
Looks good. r+

PR #213, opened 2019-10-17
Mjl index speedup

mjl:
Use elasticsearch's parallel_bulk for indexing, add ELASTICSEARCH_DSL_PARALLEL default setting and parameters to management command.
Use qs.iterator() for fetching data during reindex, as this is much more memory efficient and performant.
Instead of finding out which methods to call to prepare fields, do that finagling once and cache it for subsequent model instance prepares.

See issue #154 for performance analysis and details.

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -150,7 +150,10 @@ class DocType(DSLDocument):
     def parallel_bulk(self, actions, **kwargs):
         if self.django.queryset_pagination and 'chunk_size' not in kwargs:
             kwargs['chunk_size'] = self.django.queryset_pagination
-        deque(parallel_bulk(client=self._get_connection(), actions=actions, **kwargs), maxlen=0)
+        bulk_actions = parallel_bulk(client=self._get_connection(), actions=actions, **kwargs)
+        # As the `parallel_bulk` is lazy, we need to get it into `deque` to run it instantly
+        # See https://discuss.elastic.co/t/helpers-parallel-bulk-in-python-not-working/39498/2
+        deque(bulk_actions, maxlen=0)
         # Fake return value to emulate bulk() since we don't have a result yet,
         # the result is currently not used upstream anyway.
         return (1, [])
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -5,7 +5,6 @@ from copy import deepcopy
 from functools import partial
 
 from django import VERSION as DJANGO_VERSION
-from django.core.paginator import Paginator
 from django.db import models
 from django.utils.six import iteritems
 from elasticsearch.helpers import bulk, parallel_bulk
@@ -149,6 +148,8 @@ class DocType(DSLDocument):
         return bulk(client=self._get_connection(), actions=actions, **kwargs)
 
     def parallel_bulk(self, actions, **kwargs):
+        if self.django.queryset_pagination and 'chunk_size' not in kwargs:
+            kwargs['chunk_size'] = self.django.queryset_pagination
         deque(parallel_bulk(client=self._get_connection(), actions=actions, **kwargs), maxlen=0)
         # Fake return value to emulate bulk() since we don't have a result yet,
         # the result is currently not used upstream anyway.
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -53,8 +53,7 @@ class DocType(DSLDocument):
     def __init__(self, related_instance_to_ignore=None, **kwargs):
         super(DocType, self).__init__(**kwargs)
         self._related_instance_to_ignore = related_instance_to_ignore
-        if getattr(self._doc_type, '_prepared_fields', None) is None:
-            self._doc_type._prepared_fields = self.init_prepare()
+        self._prepared_fields = self.init_prepare()
 
     def __eq__(self, other):
         return id(self) == id(other)
@@ -75,10 +74,12 @@ class DocType(DSLDocument):
         """
         Return the queryset that should be indexed by this doc type.
         """
-        primary_key_field_name = self.django.model._meta.pk.name
-        return self.django.model._default_manager.all().order_by(primary_key_field_name)
+        return self.django.model._default_manager.all()
 
     def get_indexing_queryset(self):
+        """
+        Build queryset (iterator) for use by indexing.
+        """
         qs = self.get_queryset()
         kwargs = {}
         if self.django.queryset_pagination:
@@ -120,7 +121,7 @@ class DocType(DSLDocument):
         """
         data = {
             name: prep_func(instance)
-                for name, field, prep_func in self._doc_type._prepared_fields
+                for name, field, prep_func in self._prepared_fields
             }
         # print("-> %s" % data)
         return data
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -51,6 +51,7 @@ model_field_class_to_field_class = {
 }
 
 class DocType(DSLDocument):
+    _prepared_fields = []
     def __init__(self, related_instance_to_ignore=None, **kwargs):
         super(DocType, self).__init__(**kwargs)
         self._related_instance_to_ignore = related_instance_to_ignore
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -53,6 +53,8 @@ class DocType(DSLDocument):
     def __init__(self, related_instance_to_ignore=None, **kwargs):
         super(DocType, self).__init__(**kwargs)
         self._related_instance_to_ignore = related_instance_to_ignore
+        if getattr(self._doc_type, '_prepared_fields', None) is None:
+            self._doc_type._prepared_fields = self.init_prepare()
 
     def __eq__(self, other):
         return id(self) == id(other)
@@ -109,16 +111,13 @@ class DocType(DSLDocument):
 
             fields.append((name, field, fn))
 
-        self._doc_type._prepared_fields = fields
+        return fields
 
     def prepare(self, instance):
         """
         Take a model instance, and turn it into a dict that can be serialized
         based on the fields defined on this DocType subclass
         """
-        if getattr(self._doc_type, '_prepared_fields', None) is None:
-            self.init_prepare()
-
         data = {
             name: prep_func(instance)
                 for name, field, prep_func in self._doc_type._prepared_fields
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -125,7 +125,6 @@ class DocType(DSLDocument):
             name: prep_func(instance)
                 for name, field, prep_func in self._prepared_fields
             }
-        # print("-> %s" % data)
         return data
 
     @classmethod
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -1,9 +1,13 @@
 from __future__ import unicode_literals
 
+from collections import deque
+from copy import deepcopy
+from functools import partial
+
 from django.core.paginator import Paginator
 from django.db import models
 from django.utils.six import iteritems
-from elasticsearch.helpers import bulk
+from elasticsearch.helpers import bulk, parallel_bulk
 from elasticsearch_dsl import Document as DSLDocument
 
 from .exceptions import ModelFieldNotMappedError
@@ -45,7 +49,6 @@ model_field_class_to_field_class = {
     models.URLField: TextField,
 }
 
-
 class DocType(DSLDocument):
     def __init__(self, related_instance_to_ignore=None, **kwargs):
         super(DocType, self).__init__(**kwargs)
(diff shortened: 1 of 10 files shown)


django_elasticsearch_dsl/documents.py
```diff
@@ -124,6 +145,12 @@ def to_field(cls, field_name, model_field):
     def bulk(self, actions, **kwargs):
         return bulk(client=self._get_connection(), actions=actions, **kwargs)
 
+    def parallel_bulk(self, actions, **kwargs):
+        deque(parallel_bulk(client=self._get_connection(), actions=actions, **kwargs), maxlen=0)
```
safwanrahman: Why use `deque` here? should not `parallel_bulk` is enough?

mjl:
> Why use `deque` here? should not `parallel_bulk` is enough?

No, parallel_bulk is a generator and thus is only started when started.

I stole the "deque" method from here 
https://discuss.elastic.co/t/helpers-parallel-bulk-in-python-not-working/39498/2

safwanrahman:
@mjl Thanks for the update. I will give a closer review tomorrow and let you know the update

safwanrahman:
I was thinking how a user can configure the argument of [`parallel_bulk`](https://elasticsearch-py.readthedocs.io/en/master/helpers.html#elasticsearch.helpers.parallel_bulk) like `thread_count=4, chunk_size=500` etc. @mjl Do you have any idea?

safwanrahman:
> No, parallel_bulk is a generator and thus is only started when started.

I understand. as `parallel_bulk` works lazyly, we can use it to optimize it in more way! I will give more review tomorrow.

mjl:
I would not want to expose too many configuration options in settings, 
it just confuses 99% of the people.

If someone really really wants to override those parameters, they can 
always subclass Document and overload the parallel_bulk method to stuff 
those parameters into kwargs.

django_elasticsearch_dsl/documents.py
```diff
@@ -73,36 +76,54 @@ def get_queryset(self):
...
-                        instance, self._related_instance_to_ignore
-                    )
+                    fn = partial(field.get_value_from_instance, field_value_to_ignore=self._related_instance_to_ignore)
+
+            fields.append((name, field, fn))
 
-            data[name] = field_value
+        self._doc_type._prepared_fields = fields
 
+    def prepare(self, instance):
```
safwanrahman: Why do we need to change the prepare function?

mjl:
It calls init_prepare() on first call to construct/gather prepare functions for all fields.

mjl:
Actually, it might be better to do the init_prepare() in __init__(). Let 
me change that real quick...

safwanrahman:
The tests fails seems relative. Can you fix it?

mjl:
Oh my, I didn't expect that much breakage. I'll start mopping up right now.

mjl:
I need some guidance.

There is a test for pagination. However, pagination has been totally 
replaced with using qs.iterator(). Should I remove the test, or adapt it 
to match current query patterns, or re-add pagination on top of 
iterator() (which doesn't make a lot of sense to me and probably negates 
using iterator in the first place, but...)?

django_elasticsearch_dsl/documents.py
```diff
@@ -70,39 +76,57 @@ def get_queryset(self):
...
-    def prepare(self, instance):
+    def get_indexing_queryset(self):
         """
-        Take a model instance, and turn it into a dict that can be serialized
-        based on the fields defined on this DocType subclass
+        Build queryset (iterator) for use by indexing.
+        """
+        qs = self.get_queryset()
+        kwargs = {}
+        if DJANGO_VERSION >= (2,) and self.django.queryset_pagination:
```
safwanrahman: I think we need to handle the usecase of people who are using `django< 2` version. They should be able to paginate the queryset like before.

tests/test_integration.py
```diff
@@ -135,6 +135,7 @@ def test_get_doc_with_many_to_many_relationships(self):
         ])
 
     def test_doc_to_dict(self):
+        self.maxDiff = None  # XXX: mjl temporary
```
safwanrahman: Remove it

safwanrahman:
> I need some guidance. There is a test for pagination. However, pagination has been totally replaced with using qs.iterator(). Should I remove the test, or adapt it to match current query patterns, or re-add pagination on top of iterator() (which doesn't make a lot of sense to me and probably negates using iterator in the first place, but...)?

I think paginator can not be totally replaced if we want to keep support of `django<2.0`. So you should use paginator in the django version where `chunk_size` is not present in `iterator` and use `iterator` with `chunk_size` where its present. Regarding tests, can you fix the tests so it cover the case of iterator and Pagination both?

@mjl Will you consider adding some tests for your workflow? I know its hard to add some tests for this workflow, but I dont want things get broken in future.

mjl:
> I think paginator can not be totally replaced if we want to keep support of django<2.0.
> 
I just checked the django commit history. dj<2 does not have the chunk_size parameter but instead hardcodes it at 2000, so it's fine. dj<1.11 does not have server side cursor support in iterator so it would need manual pagination support (it should still work, but blow up with large datasets, with or without pagination -- that was part of the reason I started working on this in the first place).

Considering that official support for dj 1.10 ended in december 2017, is it worth the complexity?

safwanrahman:
> Considering that official support for dj 1.10 ended in december 2017, is it worth the complexity?

Django 1.10 is not supported, so its not needed. But django 1.11 is a important LTS release and its the last release for python 2.x. So I would consider supporting the `django 1.11` for long time. As the pagination can be configured in the document, we must support it for django 1.11 also.

mjl:
> I think paginator can not be totally replaced if we want to keep 
> support of `django<2.0`. So you should use paginator in the django 
> version where `chunk_size` is not present in `iterator`

I can wrap iterator() into another iterator for dj 1.11 that does 
chunking according to the queryset_pagination value -- would this be ok?

safwanrahman:
@mjl That seems like a good idea. Recently a PR is merged into django about having iterator in pagination. https://github.com/django/django/pull/9296

Can you copy the code from django project and extend a paginator class to support django 1.11? What do you think?

mjl:
I've looked at the django source code some more, and I'm not convinced 
that pagination of the queryset.iterator() actually _DOES_ anything. As 
I understand, django uses iterators all the way down to the database 
driver. That means:

(1) We call Document.get_indexing_queryset() which goes directly to 
Queryset.iterator(), which triggers:
(2) sql.compiler.execute_sql() reading chunk_size (either passed in from 
above layers in dj2 or hardcoded as 100 in dj1) rows from the database 
cursor. It returns an iterator that yields rows. Every time the chunk is 
exhausted, it reads the next chunk from the database.
(3) ModelIterator uses the result of sql.compiler.execute_sql() as input 
and returns an iterator that yields Model instances.
(4) Document._get_actions() takes the above and returns an iterator that 
yields the data structure for ES.
(5) ES.bulk() is consuming the iterator in its default chunk size (500).

Therefore we don't ever have more than the amount of rows 
sql.compiler.execute_sql() chunk-reads plus a couple of "in flight" rows 
being processed by the iterator pipeline in memory. Adding a "chunking 
iterator" somewhere around step 4 will do ... what exactly?



Note: It might be worthwhile to look into passing the 
queryset_pagination value to ES.bulk() at step 5, but that's beside the 
point here.

safwanrahman:
@mjl Thanks a lot for looking deep into django source code. I did not have idea about it.

Can you make a small chunk of code what you are proposing? I think passing a chunking iterator makes sense!

mjl:
> @mjl Thanks a lot for looking deep into django source code. I did not 
> have idea about it.

Me neither. It's fascinating how optimised those code paths are.

> Can you make a small chunk of code what you are proposing? I think 
> passing a chunking iterator makes sense!

   Actually, it doesn't. Really.

   I tried a 37k records test set with different settings for pagination 
size to see how much a "wrong" setting would impact indexing time.

    1         32.878s
   10         22.377s
  100         19.364s
1000         19.883s
2000         19.317s
5000         18.396s
50000        19.305s
No setting   18.802s

   So basically, it's all in the same ballpark as soon as you're out the 
pathological range.

   Therefore, I would propose to annotate the docs for 
queryset_pagination with "this parameter is ignored in Django 1.11" and 
be done with it. Or even totally deprecate it because it doesn't seem to 
have any visible impact, but that is perhaps a bit extreme an approach 
:-)

safwanrahman:
@mjl Thanks a lot for the explaination.
>Therefore, I would propose to annotate the docs for
queryset_pagination with "this parameter is ignored in Django 1.11" and
be done with it.

I am not actually very much comfortable about it. Django 1.11 is the last release that supports Python 2.x so I assume there will be a larger userbase using django 1.11 for long time. Moreover django 1.11 is still supported so we can not deprecate it. I think we need to think a little more about supporting django 1.11. Do you have any idea what can we do?

Moreover, is it possible to "Allowe edit from maintainers" so I can also help to support django 1.11?
image

mjl:
> I am not actually very much comfortable about it. Django 1.11 is the last release that supports Python 2.x so I assume there will be a larger userbase using django 1.11 for long time. Moreover django 1.11 is still supported so we can not deprecate it. I think we need to think a little more about supporting django 1.11.
> 
But that's the whole point: it __IS__ supported. It should work fine, with the only difference that the pagination parameter is not passed down into Django. And from my experiments, that parameter doesn't do much in the first place.

I'll port my test setup over to Django 1.11 and do the indexing test again to make sure indexing time is in the same ballpark.

mjl:
So the results are in. Django 1.11, DED 7.0.0, same database/dataset/psycopg2 as with the tests above.

Vanilla DED:

    $ time python manage.py search_index --rebuild -f
    Indexing 37148 'Contract' objects

    real	0m54.372s

mjl-index-speedup:

    $ time python manage.py search_index --rebuild -f
    Indexing 37148 'Contract' objects

    real	0m33.599s

So it's not as good as with Django 2, but (1) it works out of the box without any problems, and (2) is still a significant improvement over the current code base.

(Allow edit from maintainers is on BTW)

safwanrahman:
>But that's the whole point: it __IS__ supported. It should work fine, with the only difference that the pagination parameter is not passed down into Django. And from my experiments, that parameter doesn't do much in the first place.

I understand. But removing a functionality for a supported version seems like its deprecated.

So I have tested it with django 1.11 and found something interesting.
mjl-index-speedup:
**With `--parallel`**
```
time ./manage.py search_index --populate --models test_app.Car --parallel -f
Indexing 100010 'Car' objects (parallel)
       48.29 real        19.25 user         1.26 sys
```
**Without `--parallel`**
```
time ./manage.py search_index --populate --models test_app.Car -f
Indexing 100010 'Car' objects 
       24.42 real        13.14 user         0.57 sys
```

So I have overwritten the `get_indexing_queryset` method to not use `iterator`. like following:
```python
    def get_indexing_queryset(self):
        return self.get_queryset()
```

**With `--parallel`**
```
time ./manage.py search_index --populate --models test_app.Car --parallel -f
Indexing 100010 'Car' objects (parallel)
       17.20 real        14.22 user         1.34 sys
```

Without `--parallel`
```
time ./manage.py search_index --populate --models test_app.Car -f
Indexing 100010 'Car' objects 
       24.80 real        11.58 user         0.59 sys
```

So it seems like using `iterator` with `parallel` is not a good idea for django 1.11. [According to the release note, the default chunk size is `100` for 1.11. ](https://docs.djangoproject.com/en/2.2/releases/2.0/)
image
If it was 2000, I think it is [efficiant](https://www.postgresql.org/message-id/4D2F2C71.8080805%40dndg.it) for the optimization.

We also can not remove the `iterator` for django 1.11 as there can be larger dataset and the machine will be out of memory loading all the data. So we must need to keep the pagination for it. 

@mjl What do you think? Can we use the django `Pagination` for django 1.11?

mjl:
Now that is interesting, let me try that too:

```
dj 1.11 / no-parallel    33.250s
dj 1.11 / parallel       27.031s
dj 2.2 / no-parallel     27.111s
dj 2.2 / parallel        19.677s
```

So it seems that doesn't happen here... I hate non-reproducible problems. Did your objects come out of a database? Postgres or sqlite or something else?

mjl:
For giggles, I tried to patch the Django 1.11 source to use a larger chunk size in compiler.iterator(). It did NOT change anything, not even a small improvement.

> I understand. But removing a functionality for a supported version seems like its deprecated.

I think we have to agree to disagree here. In my view, it doesn't make a lot of sense to actively support something that doesn't change anything in the first place; but it's not my project at the end of the day.

> @mjl What do you think? Can we use the django `Pagination` for django 1.11?

Of course we can. I still believe it does not make any sense at all when used with iterator, but we can easily add something to get_queryset_iterator() that reads N elements and then yields them at once for Django 1.x.

mjl:
Ok, I added/adapted some tests. I fear it might be a bit brittle because it tests the inner workings of the indexing process, but hey...

django_elasticsearch_dsl/documents.py
```diff
@@ -124,6 +148,12 @@ def to_field(cls, field_name, model_field):
     def bulk(self, actions, **kwargs):
         return bulk(client=self._get_connection(), actions=actions, **kwargs)
 
+    def parallel_bulk(self, actions, **kwargs):
+        deque(parallel_bulk(client=self._get_connection(), actions=actions, **kwargs), maxlen=0)
```
safwanrahman: @mjl Can you add the `chunk_size` parameter here which will consist the value of `queryset_pagination`?

tests/test_documents.py
```diff
@@ -274,8 +274,72 @@ class Django:
...
+        class CarDocument2(DocType):
+            class Django:
+                model = Car
+
+        doc = CarDocument()
+        car1 = Car()
+        car2 = Car()
+        car3 = Car()
+        with patch('django_elasticsearch_dsl.documents.bulk') as mock_bulk, \
+             patch('django_elasticsearch_dsl.documents.parallel_bulk') as mock_parallel_bulk:
```
safwanrahman: Better to have this
```suggestion
             with (patch('django_elasticsearch_dsl.documents.bulk') as mock_bulk,
                 patch('django_elasticsearch_dsl.documents.parallel_bulk') as mock_parallel_bulk):```

File: django_elasticsearch_dsl/documents.py (hunk 5 above)
Comment (safwanrahman) [inline django_elasticsearch_dsl/documents.py]:
```suggestion
        bulk_actions = parallel_bulk(client=self._get_connection(), actions=actions,
                                         chunk_size=self.queryset_pagination, **kwargs),
        # As the `parallel_bulk` is lazy, we need to get it into `deque` to run it instantly
        deque(bulk_actions, maxlen=0)
```

safwanrahman:
@mjl Thanks a lot for the work you have done. I have added couple of reviews, after fixing this thing, I think its good to merge. 👍

mjl:
> ```suggestion
>              with (patch('django_elasticsearch_dsl.documents.bulk') as 
> mock_bulk,
>                  patch('django_elasticsearch_dsl.documents.parallel_bulk') 
> as mock_parallel_bulk):```

   Doesn't work, this isn't valid syntax according to my python 3.7. I 
incorporated the rest of your suggestions.

django_elasticsearch_dsl/documents.py
```diff
@@ -70,39 +75,57 @@ def get_queryset(self):
...
+    def prepare(self, instance):
+        """
+        Take a model instance, and turn it into a dict that can be serialized
+        based on the fields defined on this DocType subclass
+        """
+        data = {
+            name: prep_func(instance)
+                for name, field, prep_func in self._prepared_fields
+            }
+        # print("-> %s" % data)
```
safwanrahman: can you re
```suggestion

```

django_elasticsearch_dsl/documents.py
```diff
@@ -124,6 +147,17 @@ def to_field(cls, field_name, model_field):
     def bulk(self, actions, **kwargs):
         return bulk(client=self._get_connection(), actions=actions, **kwargs)
 
+    def parallel_bulk(self, actions, **kwargs):
+        if self.django.queryset_pagination and 'chunk_size' not in kwargs:
+            kwargs['chunk_size'] = self.django.queryset_pagination
```
safwanrahman: fix the indent.

safwanrahman:
> Doesn't work, this isn't valid syntax according to my python 3.7. I
incorporated the rest of your suggestions.

That is weired. Can you check how to break the line without using `\`. `\` backslach is not very much readable.

mjl:
> +    def parallel_bulk(self, actions, **kwargs):
> +        if self.django.queryset_pagination and 'chunk_size' not in kwargs:
> +            kwargs['chunk_size'] = self.django.queryset_pagination
>
> fix the indent.

  ERR_CONFUSED, I don't understand? Looks fine to me?

safwanrahman:
> ERR_CONFUSED, I don't understand? Looks fine to me?

Oh. sorry. it did not catch my eye that its a seperate block. Sorry

mjl:
> That is weired. Can you check how to break the line without using `\`. 
> `\` backslach is not very much readable.

   Agreed, but there is no way from a syntax POV. I could nest two with 
statements, but that doesn't really help much.

tests/test_documents.py
safwanrahman: What about this?
```suggestion
             bulk = "django_elasticsearch_dsl.documents.bulk"
             parallel_bulk = "django_elasticsearch_dsl.documents.parallel_bulk"
             with patch(bulk) as mock_parallel_bulk, patch(parallel_bulk) as mock_parallel_bulk:
```

pySilver:
Congratulations to you both! Thank you for your time and hard work!

mjl:
Thanks to all involved, was great working with you guys. See you on the 
next pull request! :-)

safwanrahman:
Just released a new branch `7.1.0` with the changes. https://pypi.org/project/django-elasticsearch-dsl/7.1.0/#description

mjl:
@safwanrahman The release notes description is not quite complete.
Most of the speed improvement comes from using queryset.iterator(), which also uses way less memory. Then comes precomputing prepare functions to move all those comparisons out of the criticial path; and then there is the parallel indexing.

I'm not sure that is of interest to end users to include in the release notes though. Perhaps that is too much detail.

safwanrahman:
@mjl Oops. I have updated the release note. Check it and let me know if anything need to be changed.

PR #219, opened 2019-10-31
[Fix #201] Removing old elasticsearc compatibility

### LINKED ISSUE #201: Remove old elasticsearch versions compatibility
Is there a reason for this type of compatibility code? If every version only supports the corresponding major one I don't see a reason not to remove it, specially the ES2 one.
https://github.com/sabricot/django-elasticsearch-dsl/blob/addce06166dd31a052316e07b44736b451779f75/django_elasticsearch_dsl/fields.py#L213-L252

I'll be happy to make a pull request to at least remove it from 7.x branch

safwanrahman:
This fixes  #201

### DIFF
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -1,13 +1,11 @@
 import collections
 from types import MethodType
 
-from elasticsearch_dsl.field import Keyword, Text
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 from django.db.models.fields.files import FieldFile
 from django.utils.encoding import force_text
 from django.utils.functional import Promise
-
 from elasticsearch_dsl.field import (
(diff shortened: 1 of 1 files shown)


alexgmin:
Can you also update the docs?

django_elasticsearch_dsl/fields.py
```diff
@@ -1,7 +1,7 @@
 import collections
 from types import MethodType
-import warnings
 
+from elasticsearch_dsl.field import Keyword, Text
```
alexgmin: These imports should be together with the ones in line 11.

safwanrahman: Thanks!

safwanrahman:
@alexgmin This has been updated already. in documentation it is mentioed about which version to install for which ES version.

PR #221, opened 2019-10-31
Adding documentation for parallel indexing.

safwanrahman:
Parallel indexing was implemented with https://github.com/sabricot/django-elasticsearch-dsl/pull/213 but it was not in the documentation that you can pass `parallel` command while indexing. Also, we did not prefer using `parallel` while having huge amount of data. @mjl Can have a review?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,6 +27,7 @@ Features
 - Management commands for creating, deleting, rebuilding and populating indices.
 - Elasticsearch auto mapping from django models fields.
 - Complex field type support (ObjectField, NestedField).
+- Index fast using `parallel` indexing.
 - Requirements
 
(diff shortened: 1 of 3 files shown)


mjl:
Thanks for thinking about documentation, I totally forgot to update it to the latest state. If you just accept this pull request, I'll update the missing parts and submit a new pull req for that, would that be ok?

safwanrahman:
@mjl As #222 got merged, we now have a complete documentation.(Thanks @alexgarel). I think you can pull the latest branch and add to the appropiate documentation. What do you think?

safwanrahman:
I have rebased upon master and fixed it. Seems to be good.

PR #222, opened 2019-11-01
using sphinx for documentation and publish it to readthedocs.io

alexgarel:


safwanrahman:
Thanks a lot @alexgarel for your awesome contribution.

PR #230, opened 2019-12-02
Add support for Django 3.0

noamkush:

PR #233, opened 2019-12-04
Django 3.0 support

CorrosiveKid:
Adds support for Django 3.0

Closes https://github.com/sabricot/django-elasticsearch-dsl/pull/233

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,11 +21,15 @@ matrix:
       python: 2.7
     - env: TOX_ENV=py36-django-2-es6
       python: 3.6
+    - env: TOX_ENV=py36-django-3-es6
+      python: 3.6
     - env: TOX_ENV=py37-django-2-es6
       python: 3.7
     - env: TOX_ENV=py37-django-21-es6
       python: 3.7
     - env: TOX_ENV=py37-django-22-es6
+      python: 3.7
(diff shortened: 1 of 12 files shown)


CorrosiveKid:
@sabricot @safwanrahman Please review when you have some time.

safwanrahman:
I think it was created before? https://github.com/sabricot/django-elasticsearch-dsl/pull/230

CorrosiveKid:
@safwanrahman Ah yes it's pretty much the same except this one maintains Python 2 support.

Rafalonso:
Since python 2 is in its EOL, I believe support for it should be dropped.

safwanrahman:
@Rafalonso I dont think so. We still support django 1.11. So it should be supported untill django 1.11 is supported.

Rafalonso:
@safwanrahman django 1.11 would still be supported, just in python 3. But yes, I have changed my mind, there's no reason to drop a version unless it's necessary

safwanrahman:
@Rafalonso Great. I also do not like to drop support for python2 as it will create problem for the legacy applications that need to rely on Python 2 support unless we face any issue that is hard to solve in python 2.

safwanrahman:
Thanks for the contribution. But  #230 seems to be opened before this one. So closing it.

PR #237, opened 2019-12-25
Allow choosing documents by list of index names during indexing

serkanozer:
It is possible to use multiple indices for the same model. With current `search_index` command, your only chance is to rebuild/create/delete/populate them together. However it can be desirable to run indexing separately for the different indices related to same model. This pr handles that by providing optional arg on top of `search_index` command. 

e.g.

`./manage.py search_index --rebuild index_1 index_2` will only operate on index_1 and index_2 independent of the models

### DIFF
--- a/django_elasticsearch_dsl/management/commands/search_index.py
+++ b/django_elasticsearch_dsl/management/commands/search_index.py
@@ -18,6 +18,13 @@ class Command(BaseCommand):
             help="Specify the model or app to be updated in elasticsearch"
         )
         parser.add_argument(
+            '--indices',
+            metavar='indices',
+            type=str,
+            nargs='*',
+            help="Specify the indices to be updated in elasticsearch"
+        )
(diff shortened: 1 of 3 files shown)


safwanrahman:
Can you rebase upon master and fix the merge conflict?

serkanozer:
fixed @safwanrahman

safwanrahman:
Thanks for the rebase @serkanozer. I am thinking about whether it is appropiate to implement this feature as `index name` is not unique. You can have multiple index with same name in different elasticsearch cluster. So deleting with the index name maybe raise some issues.

safwanrahman:
I will go through the pull request and think more about how to reduce the chance of duplicacy! Thanks.

safwanrahman:
This fixes #290

safwanrahman:
Currently we support indexing using model. I believe it is enough for future. So closing this.

PR #243, opened 2020-01-11
Support Elasticsearch 6

DannyAziz:
In my personal testing it seems like the only thing needed for Elasticsearch 6 to work with 7.x of this repo is:
1. To be using `elasticsearch>=6.0.0,<7.0.0 elasticsearch-dsl>=6.0.0,<7.0.0`
2. For the object sent to elasticsearch during indexing includes `_type`

I've implemented the ability to determine if you are using ES6 if so then the `_type` field is included

This *should* work with both ES6 and ES7, haven't tested on any other versions

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -49,7 +49,6 @@ model_field_class_to_field_class = {
     models.URLField: TextField,
 }
 
-es_connection = get_connection()
 
 class DocType(DSLDocument):
     _prepared_fields = []
@@ -168,7 +167,7 @@ class DocType(DSLDocument):
                 self.prepare(object_instance) if action != 'delete' else None
             ),
         }
-
+        es_connection = get_connection()
         if es_connection.info()['version']['number'][0] == '6':
             prepared_object['_type'] = self._doc_type.name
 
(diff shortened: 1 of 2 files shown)


django_elasticsearch_dsl/documents.py
```diff
@@ -157,14 +159,19 @@ def parallel_bulk(self, actions, **kwargs):
...
+        prepared_object = {
             '_op_type': action,
             '_index': self._index._name,
             '_id': object_instance.pk,
             '_source': (
                 self.prepare(object_instance) if action != 'delete' else None
             ),
         }
+        es_connection = get_connection()
+        if es_connection.info()['version']['number'][0] == '6':
```
safwanrahman: I think it is not proper as it will call to elasticsearch for knowing the version. So the indexing will be pretty slow!

barseghyanartur: Agree. There's [another PR](https://github.com/django-es/django-elasticsearch-dsl/pull/288) addressing this issue, which includes tests, travis changes and convenient way to detect which version of Elasticsearch is used.

safwanrahman:
We have created a separate branch for 6.x things. So closing this!

Issue #155, opened 2019-01-30
User as SimpleLazyObject not possible to index


ruddy2442:
Hi, I am having troubles indexing the User when user object has been got from request (and authenticated over default modelbackend auth). Django uses its SimpleLazyObject and django dsl crashes somewhere in  django_elasticsearch_dsl/fields.py:get_value_from_instance() on line 127 with error {TypeError} 'User' object is not iterable. When user is passed to indexer as User from DB, indexing works, when its passed from request.user, it does not work (thats the case, when we are creating object with relation to user and user is passed to create method as request.user - SimpleLazyObject). Than the signal is called on instance to index, but objects is still the SimpleLazyObject, instead of user.

abdelhalimresu:
@ruddy2442, I believe it's the same issue I was facing.
I resolved it by using the `request.user.id` instead of `request.user` when creating an object which depends on the user object like this:
```python
object = YourModel(**kwargs, user_id=request.user.id)
object.save()
```
instead of
```python
object = YourModel(**kwargs, user=request.user)
object.save()
```

manishkumr:
I am exactly having the same issue while trying to save a model which has User model as a foreign key.  I have got this working as suggested by @ruddy2442 , will there be any fix available for this?

mboudet:
Same issue here. In one of my model, I manually set the created_by field to the user making the request. The field is populated with request.user in form_valid.

Indexing an existing entity works fine, but creating a new one crashes with "'User' object is not iterable"

A workaround was to use get_user(request) to get the 'real' user and not the lazy object.

While it isn't much trouble, the error thrown by django-elasticsearch-dsl is not very clear. it might be good to add to the documentation that 'request.user' cannot be used by itself at entity creation?
(And all SimpleLazyObjects)

hemache:
A quick workaround would be to fix `ObjectField` by handling `SimpleLazyObject`s gracefully 

```
from django.utils.functional import SimpleLazyObject
from django_elasticsearch_dsl import fields

class MyObjectField(fields.ObjectField):
    def get_value_from_instance(self, instance, field_value_to_ignore=None):
        objs = fields.DEDField.get_value_from_instance(self, instance, field_value_to_ignore)

        if objs is None:
            return {}
        if not isinstance(objs, SimpleLazyObject) and isinstance(objs, collections.Iterable):
            return [
[...]

PR #255, opened 2020-03-10
Handle case where SimpleLazyObject being treated as an Iterable

hemache:
fixes https://github.com/sabricot/django-elasticsearch-dsl/issues/155

### DIFF
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -5,7 +5,7 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 from django.db.models.fields.files import FieldFile
 from django.utils.encoding import force_text
-from django.utils.functional import Promise, SimpleLazyObject
+from django.utils.functional import Promise
 from elasticsearch_dsl.field import (
     Boolean,
     Byte,
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -124,7 +124,12 @@ class ObjectField(DEDField, Object):
 
         if objs is None:
             return {}
-        if not isinstance(objs, SimpleLazyObject) and isinstance(objs, collections.Iterable):
+        try:
+            is_iterable = bool(iter(objs))
+        except TypeError:
+            is_iterable = False
+        
+        if is_iterable:
             return [
                 self._get_inner_field_data(obj, field_value_to_ignore)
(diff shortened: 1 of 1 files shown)


django_elasticsearch_dsl/fields.py
```diff
@@ -124,7 +124,7 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
 
         if objs is None:
             return {}
-        if isinstance(objs, collections.Iterable):
+        if not isinstance(objs, SimpleLazyObject) and isinstance(objs, collections.Iterable):
```
safwanrahman: I think it can be better written as following so it is more pythonic and readable

```suggestion
       objs_type = type(objs)

        if objs_type is collections.Iterable and objs_type is not SimpleLazyObject:
```

safwanrahman:
@hemache Thanks for your PR. I am thinking about the situation where the SimpleLazyObject is a ctually a iterable, in the scenario, it will not work.

django_elasticsearch_dsl/fields.py
safwanrahman: This should be better approach as the `SimpleLazyObject` can also be a `Iterable`.
```suggestion
        try:
            is_iterable = bool(iter(objs))
        except TypeError:
            is_iterable = False
        
        if is_iterable:
```

safwanrahman:
@hemache Sorry for asking you, will you have some time to make the changes?

fabio-filho:
@hemache, do you need some help ?

hemache:
hi guy, I'm really sorry I didn't have time to work on this lately, I would appreciate any help you could provide

hemache:
@safwanrahman you're right, I accepted your suggested changes.

django_elasticsearch_dsl/fields.py
```diff
@@ -5,7 +5,7 @@
 from django.db import models
 from django.db.models.fields.files import FieldFile
 from django.utils.encoding import force_text
-from django.utils.functional import Promise
+from django.utils.functional import Promise, SimpleLazyObject
```
safwanrahman, line 8:
```suggestion
from django.utils.functional import Promise
```
One small nit, need to remove importing SimpleLazyObject

andreasnuesslein:
Also, first line: `import collections` is unnecessary now.

safwanrahman:
LGTM. r+

PR #262, opened 2020-04-16
Create mappings for multiple document types in one index

prebenkeeeb:
With the removal of types in ES6 there is a need to create a custom type in order to search only one index as described in:
[removal-of-types](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html)

The problem in the current implementation with multiple document types using the same index is that only the mappings of the first registered document type is applied when calling "python manage.py search_index --create". 
  
This pull request merges the mappings from the other document types(which use the same index) onto the index of the first registered document type.  

There was also work done in elasticsearch-dsl to merge mappings of multiple docTypes in the same index - see: 
[elasticsearch-dsl-py/issues/800](https://github.com/elastic/elasticsearch-dsl-py/issues/800)
and changes here:
[(https://github.com/elastic/elasticsearch-dsl-py/commit/11db37586b897a9313213d3bbd2c55f9ba20a90c](https://github.com/elastic/elasticsearch-dsl-py/commit/11db37586b897a9313213d3bbd2c55f9ba20a90c)

### DIFF
--- a/django_elasticsearch_dsl/registries.py
+++ b/django_elasticsearch_dsl/registries.py
@@ -31,6 +31,8 @@ class DocumentRegistry(object):
         for idx, docs in iteritems(self._indices):
             if index._name == idx._name:
                 docs.add(doc_class)
+                for doc_types in index._doc_types:
+                    idx._doc_types.append(doc_types)
                 return
 
         self._indices[index].add(doc_class)
--- a/django_elasticsearch_dsl/registries.py
+++ b/django_elasticsearch_dsl/registries.py
@@ -31,9 +31,9 @@ class DocumentRegistry(object):
         for idx, docs in iteritems(self._indices):
             if index._name == idx._name:
                 docs.add(doc_class)
-                for doc_types in index._doc_types:
-                    idx._doc_types.append(doc_types)
-                return
+                if doc_class not in idx._doc_types:
+                    idx._doc_types.append(doc_class)
+                    return
 
         self._indices[index].add(doc_class)
 
--- a/tests/test_documents.py
+++ b/tests/test_documents.py
@@ -359,7 +359,6 @@ class DocTypeTestCase(TestCase):
                          [('name', (), {}),  ('price', (), {}), ('type', (), {})]
         )
 
-
     def test_multiple_docs_in_same_index(self):
         car_doc = CarDocument()
         car_doc_mappings = car_doc._index.to_dict().get('mappings')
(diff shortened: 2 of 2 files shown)


prebenkeeeb:
Ping!

safwanrahman:
@prebenkeeeb Thanks for the PR. Can you explain about the use case? I can not find why you will have multiple document for same index.

prebenkeeeb:
@safwanrahman It is described here https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html#_alternatives_to_mapping_types
In summary: With multiple documents in the same index you do not waste primary shards on document types of small size and the query is simpler as you only query one index.

safwanrahman:
I am not sure actually if this is in the scope of this project actually. Moreover, I am not so comfortable to use a internal API `_doc_types` which is not documented anywhere and the use case of it. I will try to reach out some people to know more about this.

safwanrahman:
I think you can achieve the functionality in vanilla elasticsearch-dsl feature.
Something like following
```python
from elasticsearch_dsl import Index
common_index = Index("common_index")

@registry.register_document
@common_index.document
class BicycleDocument(DocType):
    color = fields.TextField()
    type = fields.TextField()

@registry.register_document
@common_index.document
class CarDocument(Document):
    class Django:
        model = Car
        fields = [
            'name',
            'color',
        ]

```

I believe this is the best way to implement the feature rathar than changing the mutable `index` object from our package. It may lead to uncertain behavior. What do you think @prebenkeeeb ?

alexgarel:
@safwanrahman, on my side I think the use case is ok, as it is well supported by Elasticsearch, and it corresponds to a real need. (It will be needed if someone wants to do percolation for example, as requests has to be in the same index as documents).

I think manipulating `_doc_types` is not a problem. We are integrating a framework to another, we are not mere users. So it is totally justified to go into its guts from time to time.

safwanrahman:
@alexgarel I understand your point. But I was thinking the functionality could be achieved using the approach I mentioned above and it is more explicit using the native way. Moreover, choosing a index based on the name is not a proper way as the can be same index name but using different cluster.

prebenkeeeb:
@safwanrahman Thanks for your reply! I was able to create multiple document types in one index with  the vanilla elasticsearch-dsl feature you suggested 👍 I believe that this is the preferred way so feel free to close this PR.

safwanrahman:
@prebenkeeeb Thanks! Closing this as we agree that the best way to solve it is using the upstream package features.

Issue #271, opened 2020-05-04
Configure Elasticsearch _id dynamically from document


jfschaff:
Hi,

I'm using `django-elasticsearch-dsl` to index two different Django models into a single ES index.

To get a common unique id for both models I use a custom `id` field specially generated for my ES index (I could not rely on individual model ids, as I may have two objects with the same id, or the ids could even be of different types).

The code looks like:
```python
class ModelA(Model):
    [...]


class ModelB(Model):
    [...]


@registry.register_document
@common_index.document
class AToCommonDocument(Document):
    class Django:
        model = ModelA
    id = fields.KeywordField()
    def prepare_id(self, instance):
        return <some_common_id_for_A_and_B>
    [...]
 

@registry.register_document
@common_index.document
class BToCommonDocument(Document):
    class Django:
        model = ModelB
    id = fields.KeywordField()
    def prepare_id(self, instance):
        return <some_common_id_for_A_and_B>
    [...]
```
This works well, and allows me to have a single index and search it efficiently, even if the underlying objects actually come from two tables. The ES entries look like:
```
{
    "_index" : "common",
    "_type" : "_doc",
    "_id" : "2",
    "_score" : 1.0,
    "_source" : {
        "id" : 110,
        [other fields]
    }
}
```
Now, my issue is that I sometimes get two documents with the same ES id (`_id`), so some documents are lost, because they are overwritten when a second, with identical `_id`, is written. I also noticed that the ES `_id` does not have the same value as the `id` field as defined in the documents above.

So the questions:
1. how is the ES `_id` determined if it is neither the Django id, not the id defined in the document? How does an update works, so that the right document is overwritten?
2. is there a way to control the ES `_id` field? For instance, setting `id` and `_id` to be the same?

safwanrahman:
As per the source [code](https://github.com/sabricot/django-elasticsearch-dsl/blob/master/django_elasticsearch_dsl/documents.py#L159-L167), the package is generating the elasticsearch `_id` from the object `pk`.

You can overwrite the method `_prepare_action` in your document and change the `_id` as per your need.
Keep in mind that as this is a internal method and not documented, it can be changed in the future version of this package. Better to write proper test for that.

jfschaff:
@safwanrahman  do you think it would make sens to add a feature for finer control of the ES `_id`, and which would default to the current behavior (object's `pk`)?

[...]

PR #272, opened 2020-05-07
Configure Elasticsearch _id dynamically from document 

jfschaff:
Implements #271

CL:
- added optional method `document_id(self, instance)` to `Document` class to customize Elasticsearch document ids
- documented that in sections `fields`

I wasn't sure where to put the doc, so don't hesitate to move it around it you think it shouldn't be in this section.

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -156,11 +156,19 @@ class DocType(DSLDocument):
         # the result is currently not used upstream anyway.
         return (1, [])
 
+    def document_id(self, object_instance):
+        """
+        The default behavior is to use the Django object's pk (id) as the 
+        elasticseach index id (_id). If needed, this method can be overloaded 
+        to change this default behavior.
+        """
+        return object_instance.pk
+
     def _prepare_action(self, object_instance, action):
         return {
             '_op_type': action,
             '_index': self._index._name,
-            '_id': object_instance.pk,
+            '_id': self.document_id(object_instance),
             '_source': (
                 self.prepare(object_instance) if action != 'delete' else None
             ),
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -156,7 +156,8 @@ class DocType(DSLDocument):
         # the result is currently not used upstream anyway.
         return (1, [])
 
-    def generate_id(self, object_instance):
+    @classmethod
+    def generate_id(cls, object_instance):
         """
         The default behavior is to use the Django object's pk (id) as the 
         elasticseach index id (_id). If needed, this method can be overloaded 
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -156,7 +156,7 @@ class DocType(DSLDocument):
         # the result is currently not used upstream anyway.
         return (1, [])
 
-    def document_id(self, object_instance):
+    def generate_id(self, object_instance):
         """
         The default behavior is to use the Django object's pk (id) as the 
         elasticseach index id (_id). If needed, this method can be overloaded 
@@ -168,7 +168,7 @@ class DocType(DSLDocument):
         return {
             '_op_type': action,
             '_index': self._index._name,
-            '_id': self.document_id(object_instance),
+            '_id': self.generate_id(object_instance),
             '_source': (
                 self.prepare(object_instance) if action != 'delete' else None
             ),
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -369,7 +369,9 @@ class IntegrationTestCase(ESTestCase, TestCase):
         article.save()
 
         # assert that the document's id is the id of the Django object
-        try:
+        es_obj = ArticleDocument.get(id=obj_id)
+        self.assertEqual(es_obj._id, article.id)
+        self.assertEqual(es_obj.slug, article.slug)
             ArticleDocument.get(id=obj_id)
         except NotFoundError:
             self.fail("document with _id {} not found").format(obj_id)
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -369,12 +369,11 @@ class IntegrationTestCase(ESTestCase, TestCase):
         article.save()
 
         # assert that the document's id is the id of the Django object
-        es_obj = ArticleDocument.get(id=obj_id)
-        self.assertEqual(es_obj._id, article.id)
-        self.assertEqual(es_obj.slug, article.slug)
-            ArticleDocument.get(id=obj_id)
+        try:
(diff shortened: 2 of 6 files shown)


django_elasticsearch_dsl/documents.py
```diff
@@ -156,11 +156,19 @@ def parallel_bulk(self, actions, **kwargs):
         # the result is currently not used upstream anyway.
         return (1, [])
 
+    def document_id(self, object_instance):
```
safwanrahman, line 159: Can you change the method name to `get_document_id`. `document_id` seems like a `property` from developer point of view.

safwanrahman: Can you also add a test for this?

safwanrahman:
@jfschaff Thanks a lot for the PR. I have added 2 improvements which I think are helpful. Let me know if you have concerns.

django_elasticsearch_dsl/documents.py
jfschaff: I get your point concerning the method name, but I have a feeling that `get_document_id` may be misleading, as this is not really a getter. It's more like a setter. It is basically the same as the `prepare_<field name>` methods, but for the document id instead of one of its fields.

I did not use the name `prepare_document_id` or `prepare_id`, as this would conflict with the methods that would set the `document_id` or `id` field inside the document.

Would you be ok with one of the following names?
- `set_id`
- `set_document_id`
- `determine_document_id`
- `determine_id`

Tell me which one you prefer and I'll do the changes.

And I'll add tests as well.

safwanrahman: I think it is not setter as it is not setting any atribute to the object itself. the method is returning the id so from independent prespective, it is a getter. You can change it to something like `generate_id`, which maybe more explicit.

jfschaff: Yes, you're right, it's not a setter either. Ok for `generate_id` then.

django_elasticsearch_dsl/documents.py
```diff
@@ -156,11 +156,19 @@ def parallel_bulk(self, actions, **kwargs):
         # the result is currently not used upstream anyway.
         return (1, [])
 
+    def generate_id(self, object_instance):
```
safwanrahman, line 159: It is better to have static method so it can be called directly without initiating the object. Like `Document.generate_id(model_object)`

safwanrahman:
I will check the tests and add some more review.

jfschaff:
OK, then I wait for your review before I make the final fixes.

tests/test_integration.py
```diff
@@ -352,3 +355,41 @@ def test_queryset_iterator_queries(self):
...
+            slug=article_slug,
+        )
+
+        # saving should create two documents (in the two indices): one with the 
+        # Django object's id as the ES doc _id, and the other with the slug 
+        # as the ES _id
+        article.save()
+
+        # assert that the document's id is the id of the Django object
+        try:
```
safwanrahman, line 372: I think it is better to check the slug as well so we can be confirm that the object is same as that.
Something like following.
```suggestion
        es_obj = ArticleDocument.get(id=obj_id)
        self.assertEqual(es_obj._id, article.id)
        self.assertEqual(es_obj.slug, article.slug)
```

tests/test_integration.py
```diff
@@ -352,3 +355,41 @@ def test_queryset_iterator_queries(self):
...
+            slug=article_slug,
+        )
+
+        # saving should create two documents (in the two indices): one with the 
+        # Django object's id as the ES doc _id, and the other with the slug 
+        # as the ES _id
+        article.save()
+
+        # assert that the document's id is its the slug
+        try:
```
safwanrahman, line 389: Same as here
```suggestion
        es_obj = ArticleDocument.get(id=article_slug)
        self.assertEqual(es_obj._id, article.article_slug)
        self.assertEqual(es_obj.slug, article.slug)
```

safwanrahman:
@jfschaff The tests looks pretty good. Can you add a unittest in the [`tests_document.py`](https://github.com/sabricot/django-elasticsearch-dsl/blob/master/tests/test_documents.py#L59)? Which will test whether the `generate_id` is called while indexing as well as the `generate_id` is returning correct value by default.

safwanrahman:
@jfschaff I was thinking about adding a configuration in the `class Django` named something like `id_field_name` which can be set to easily configure the Elastic ID without overwriting the `generate_id` method. So in the `generate_id` method, we can call `getattr(obj, self.django.id_field_name)` and return the value. What do you think?

It is not necessary to have this in this patch, we can implement that in separate PR. But are you interested to add that as well in this PR?

jfschaff:
Hi, and thank you for your feedback.

I could not find a way to retrieve the ES `_id` of a document (what you wrote `es_obj._id` in your suggestion).

Therefore, the integration tests uses to `try / except` block to get a doc by its ES `_id`. The idea of using a `try / except` is to "assert that the code does not error". I could not find another way to assert that and it looks like `assertDoesNotError` is kinda missing (see e.g. [this SO discussion](https://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises)). Doing the same without a `try / except` would error instead of failing, which would be ambiguous: we want the test to fail when the feature is broken, not to error.

As for the other points:
- I turned the method into a class method, which makes sense. A static method would not work ;
- I've tried to add unit tests, but I could not make it work. I wanted to check that `generate_id` is called but did not manage to. I'm not so used to the mock module... **I've left the commented (broken) tests at the very end of `test_documents.py` if you want to have a look at it.**

And yes, your suggestion about adding `id_field_name` is relevant. I did not think of that because it would not help for my use case. But can we first merge this PR before adding that? You can attribute that feature to me so I don't forget :-)

nicholasamorim:
Are we able to merge this in?

safwanrahman:
@nicholasamorim @jfschaff Sorry I could not manage time to fix the tests. I will look on this upcoming weekend for sure.

safwanrahman:
@jfschaff Thanks for your patience. I have fixed the tests in my repo https://github.com/safwanrahman/django-elasticsearch-dsl/commit/4c88715d can you fetch and cherry-pick the commit and push this to your branch? Or you may mark the ["allow edit from maintainers"](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) so I can push the changes directly to your repo.

The test fix is quite interesting and complex. The method is not called untill the low level elasticsearch API execute it. So the tests was not picking up.

jfschaff:
@safwanrahman, I've just pushed your commit to my branch, but I don't see the tests running...

safwanrahman:
Seems like the tests got passed! Going to merge this! Thanks a lot!

PR #284, opened 2020-07-16
Allow custom routing and index prefixing during bulk indexing using aliases and index templates.

martyzz1:


safwanrahman:
Thanks for the PR. Can you add some documentation for this?

martyzz1:
I've pushed up some more changes. this now seems to work fine allowing creation of indexes and aliases.

What I'm not sure about is how to deal with the read from index...   I'm in a microservices environment where my rread and write are on separate services so its not a problem for me to specify the write index as "thing" and on the other service specify the read index as "thing_r_alias"

martyzz1:
With regards to documentation The new management command es_reindex.py Makes use of all the features.. Not sure where to document it all?

safwanrahman:
@alexgarel Can you help him to put the documentations?

timo-piirmann:
👍  for this effort. Looks exactly what I need to replace my custom scripts in the system.

violuke:
Will this also resolve? #75 If so, is there anything holding this up other than documentation, I'd be happy to help out as #75 would be really helpful for us?

martyzz1:
I think it would...

safwanrahman:
@martyzz1 I think you can document it in the management command page. Let me know if you need any help documenting it.

safwanrahman:
Looks like it is implemented with  #358

PR #301, opened 2020-10-31
[Fix #111] Adding API for filtering object while indexing

### LINKED ISSUE #111: Updating index do not respect get_queryset method
While populating ES index, it calls `get_queryset` from the `DocType`. So if I overwrite the `get_queryset` method and filter the queryset to index only specific kind of object, it works perfectly only when I run from the management command (`rebuild` or `populate`)

But, while new objects get created it does not respect that queryset. So though I add a filter in the `get_queryset` method, all objects get indexed.

Below is the example

```python
# models.py
class Foo(model.Model):
    name = models.CharField(max_length=100)
    age = models.IntegerField()

# documents.py
@foo_index.doc_type
class FooDocument(DocType):
    class Meta:
         model = Foo
         fields = ('name', 'age')

     def get_queryset(self):
         queryset = super().get_queryset()
         return queryset.filter(age__gt=10)
```

So while running `./manage.py search_index --rebuild`, it indexes only the object that have age more than 10. But If we add another `Foo` object with `age=9`, it also get indexed.

safwanrahman:
This fixes #111. We need to add documentation for this. But I am not sure where to add it.
overwriting the `should_index_object` and returning a boolean value should determine which object should be index.

For example, if we want to index Car only with Red `color`, we can write following
```python
class Car(Document):
    def should_index_object(self, obj):
        if obj.color == "RED":
            return True
        else:
            return False
```

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -189,6 +189,7 @@ class DocType(DSLDocument):
             return self.bulk(*args, **kwargs)
 
     def should_index_object(self, obj):
+        """Overwriting this method and returning a boolean value should determine whether the object should be index."""
         return True
 
     def update(self, thing, refresh=None, action='index', parallel=False, **kwargs):
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -159,8 +159,8 @@ class DocType(DSLDocument):
     @classmethod
     def generate_id(cls, object_instance):
         """
-        The default behavior is to use the Django object's pk (id) as the 
-        elasticseach index id (_id). If needed, this method can be overloaded 
+        The default behavior is to use the Django object's pk (id) as the
+        elasticseach index id (_id). If needed, this method can be overloaded
         to change this default behavior.
         """
         return object_instance.pk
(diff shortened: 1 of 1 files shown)


safwanrahman:
@barseghyanartur @alexgarel Any thoughts about the code and where should we document it? any suggestion about the API or anything?

barseghyanartur:
@safwanrahman:

What about in here: [es_index](https://github.com/django-es/django-elasticsearch-dsl/blob/master/docs/source/es_index.rst)? Seems quite logical to me.

safwanrahman:
@barseghyanartur I was also thinking same. Lets see what @alexgarel thinks about this, as he is good at documentations.

alexgarel:
> What about in here: [es_index](https://github.com/django-es/django-elasticsearch-dsl/blob/master/docs/source/es_index.rst)? Seems quite logical to me.

I think this is a good idea ! It's not that at this point documentation is in a ideal state, but better put this information somewhere and move it latter, if needed.

Also put a docstring on `should_index_object` explaining the feature. (a lot of us look directly in the code + it would autodocument when we setup sphinx to do so)

willbarton:
We've discovered the bug described in #111 and were looking at potential fixes and work-arounds. Is this PR just waiting on documentation?

safwanrahman:
@willbarton Yes. it is waiting for the documentation. It will be great help if anyone can write something up!

barseghyanartur:
@safwanrahman:

Shall I just do this? And it could be improved later on (if needed).

> @safwanrahman:
> 
> What about in here: [es_index](https://github.com/django-es/django-elasticsearch-dsl/blob/master/docs/source/es_index.rst)? Seems quite logical to me.

safwanrahman:
@barseghyanartur Sure! 🚀

safwanrahman:
Seems like we need to merge this rathar than waiting for documentation! Added docstring though! 😅

Issue #235, opened 2019-12-21
mutable content of an Object field | index population from dynamic json


titovanton:
Hello, may be I do not understand something, but hope there is a solution for my case. I have few fields, which I've specified in the Document. One of them I've set as `data = fields.ObjectField(dynamic=True)` with the purpose to store there scheema-less object data. If I POST such documents with Kibana, it works and _search lookup inside as well, giving me documents I'm searching. But doing the same via `django-elasticsearch-dsl`, I have data field like that: 
```         
 "data" : [
            { },
            { },
            { },
            { }
          ],
```
Firts, the "data" expected to be an object, not a list. Second, it is the list of empty(!) objects.
What is the best way to implement such functionality? I can not specify the list of properties for the Object field, while the content of data field is mutable.

```
# models.py

class Car(models.Model):
    TYPE_SEDAN = 1
    TYPE_TRUCK = 2
    TYPE_SUV = 4
    TYPE_CHOICES = [
        (TYPE_SEDAN, 'Sedan'),
        (TYPE_TRUCK, 'Truck'),
        (TYPE_SUV, 'SUV'),
    ]

    name = models.CharField(max_length=256)
    color = models.CharField(max_length=256)
    description = models.TextField()
    type = models.IntegerField(choices=TYPE_CHOICES, default=1)
    data = JSONField(default=dict)

    def type_to_string(self):
        return dict(self.TYPE_CHOICES)[self.type]
```
```
# documents.py

@registry.register_document
class CarDocument(Document):
    type = fields.TextField(attr='type_to_string')
    data = fields.ObjectField(dynamic=True)

    class Index:
        # Name of the Elasticsearch index
        name = 'cars'
        # See Elasticsearch Indices API reference for available settings
        settings = {'number_of_shards': 1,
                    'number_of_replicas': 0}

    class Django:
        model = Car

        fields = [
            'name',
            'color',
            'description',
        ]
```

```
# population management command

class Command(BaseCommand):
    def handle(self, *args, **options):
        fake = Faker()
        for x in range(10):
            Car.objects.create(
                name=fake.first_name,
                color=fake.safe_color_name(),
                description='with data dyn',
                type=random.choice(list(dict(Car.TYPE_CHOICES).keys())),
                data={
                    'country': fake.country(),
                    'contact_person': fake.first_name(),
                    'info': {'word': fake.word()},
                    'items': [{'name': fake.company(), 'address': fake.address()} for x in range(5)]
                }
[...]

Issue #36, opened 2017-07-04
Support for JSONField


MightySCollins:
Hi

I am trying to use the Django models JSONField but because its structure can be changed I am not sure of its properties. Is there anyway I can avoid setting them when using a ObjectField? Or can you think of any other solutions?

Without setting properties like below the following happens
```python
attributes = fields.ObjectField()
```
```json
"attributes": [
  {
  },
  {
  }
],
```

You package has saved me so much time, thanks.

boccheciampe:
Hello,

I recently ran into the same issue. A (dirty ?) workaround is to use the prepare_foo function, something like this (in this example, "yourfield" is a relationship object with regular fields and a 'data' jsonfield):

    class MyIndex(DocType):
        [...]

        yourfield = fields.ObjectField()

        def prepare_yourfield(self, instance):
            json_keys = [list_of_all_possible_keys_of_the_jsonfield]
            final_list = []

            for yourfield in instance.yourfields.all():
                # regular fields values
                fields_dict = {
                    'attr1': yourfield.attr1,
                    'attr2: yourfield.attr2,
                    [...]
                }
                # flatten the json field values
                for k in keys:
                    fields_dict[k] = yourfield.data.get(k)

                final_list.append(fields_dict)
        return final_list

Hope this helps :)

cocoakekeyu:
i try this:
```python
class MyType(DocType):
    content_json = fields.ObjectField()

    def prepare_content_json(self, instance):
        return instance.content_json

```

andreasnuesslein:
so, part of it is in master:

https://github.com/sabricot/django-elasticsearch-dsl/pull/220

Allan-Nava:
> i try this:
> 
> ```python
> class MyType(DocType):
>     content_json = fields.ObjectField()
> 
>     def prepare_content_json(self, instance):
>         return instance.content_json
> ```

it works this solution?

andreasnuesslein:
> it works this solution?

? yes? did you try it? people _are_ giving a thumbs up

debnet:
It works, but for few records only. I want to index about 10 millions records and refresh command takes all my memory and finally crashes. I hope for a real fix because JSONField are ratherly common now in PostgreSQL.

Issue #13, opened 2017-03-30
Using Objectfield- VariableLookupError for self


yarnball:
Hi,

I am trying to return data from a function, but keep getting a "VariableLookupError" for my function `for key [self]`.

I have tried both the `NestedField` and `ObjectField` options. Both gave the same traceback (as per below).

How can I fix the lookup error?

# documents.py
```
vehicle = Index('vehicle')

vehicle.settings(
    number_of_shards=1,
    number_of_replicas=0
)

@vehicle.doc_type
class VehicalDocument(DocType):
    tag = fields.ObjectField(attr="self.get_grouped_tags")
    class Meta:
        model = Vehicle
        fields = [
            'name'
        ] 
```

# models.py
```
class Tag(models.Model):
    name = models.CharField("Name", max_length=5000, blank=True)
    taglevel = models.IntegerField("Tag level", null=True, blank=True)

class Vehicle(models.Model):
    title = models.CharField("Title", max_length=10000, blank=True)
    tag = models.ManyToManyField('Tag', blank=True)

    def get_grouped_tags(self):
        tag = self.tag.order_by('taglevel')
        grouped_tags = {
            tag_level: [
                { 'name': tag_of_level.name, 'taglevel': tag_of_level.taglevel, }
                for tag_of_level in tags_of_level
            ] for tag_level, tags_of_level
            in groupby(tag, lambda tag: tag.taglevel)
        }
        return grouped_tags
```

# Traceback
```
Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/UserX/Dev/A51/Backend/Chat/venv/lib/python2.7/site-packages/django/core/management/__init__.py", line 350, in execute_from_command_line
    utility.execute()
  File "/Users/UserX/Dev/A51/Backend/Chat/venv/lib/python2.7/site-packages/django/core/management/__init__.py", line 342, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/UserX/Dev/A51/Backend/Chat/venv/lib/python2.7/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/UserX/Dev/A51/Backend/Chat/venv/lib/python2.7/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
[...]

PR #302, opened 2020-10-31
Updated ObjectField to allow for unmapped keys

oehrlein:
I adjusted the ObjectField class to allow for data with unmapped keys. Previously, if an ObjectField (or NestedField) was created without mapping, it'd show up empty in the index.

For example:
```
# Source data
{"make": "BMW", "model": "M5"}

# documents.py
class Car(Document):
    data = fields.ObjectField()

# Index before fix
"_source" : {
    "data": [
        {}
    ]
}

# Index after fix
"_source" : {
    "data": {
        "make": "BMW",
        "model": "M5"
    }
}
```

This prevents the need for the workaround mentioned in #36, which uses a `prepare_` function such as this:
```
def prepare_data(self, instance):
    return instance.data
```

This fixes issues highlighted in #13, #36, and #235.

### DIFF
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -115,6 +115,9 @@ class ObjectField(DEDField, Object):
                     obj, field_value_to_ignore
                 )
 
+        if not data and obj and isinstance(obj, dict):
+            data = obj
+
         return data
 
     def get_value_from_instance(self, instance, field_value_to_ignore=None):
@@ -125,7 +128,7 @@ class ObjectField(DEDField, Object):
         if objs is None:
             return {}
         try:
-            is_iterable = bool(iter(objs))
+            is_iterable = bool(iter(objs)) and not isinstance(objs, dict)
         except TypeError:
             is_iterable = False
         
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -131,7 +131,7 @@ class ObjectField(DEDField, Object):
             is_iterable = bool(iter(objs))
         except TypeError:
             is_iterable = False
-        
+
         if is_iterable and not isinstance(objs, dict):
             return [
                 self._get_inner_field_data(obj, field_value_to_ignore)
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -128,11 +128,11 @@ class ObjectField(DEDField, Object):
         if objs is None:
             return {}
         try:
-            is_iterable = bool(iter(objs)) and not isinstance(objs, dict)
+            is_iterable = bool(iter(objs))
         except TypeError:
             is_iterable = False
         
-        if is_iterable:
+        if is_iterable and not isinstance(objs, dict):
             return [
                 self._get_inner_field_data(obj, field_value_to_ignore)
                 for obj in objs if obj != field_value_to_ignore
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -115,6 +115,8 @@ class ObjectField(DEDField, Object):
                     obj, field_value_to_ignore
                 )
 
+        # This allows for ObjectFields to be indexed from dicts with
+        # dynamic keys (i.e. keys/fields not defined in 'properties')
         if not data and obj and isinstance(obj, dict):
             data = obj
 
@@ -132,6 +134,8 @@ class ObjectField(DEDField, Object):
         except TypeError:
             is_iterable = False
 
+        # While dicts are iterable, they need to be excluded here so
+        # their full data is indexed
         if is_iterable and not isinstance(objs, dict):
             return [
                 self._get_inner_field_data(obj, field_value_to_ignore)
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -100,6 +100,36 @@ class ObjectFieldTestCase(TestCase):
             'last_name': "bar",
         })
 
+    def test_get_value_from_instance_with_partial_properties(self):
+        field = ObjectField(
+            attr='person',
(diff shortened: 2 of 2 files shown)


safwanrahman:
LGTM. @oehrlein Is it possible for you to add some tests? if not possible, we can merge as it is.

safwanrahman:
@oehrlein BTW, Thanks for your PR. I was just trying to figure out what was happening here!

oehrlein:
@safwanrahman Re: tests, I can take a crack at it, but not sure when I'd be able to get it done. I could potentially try to start working on it later today, but I can't make any guarantees. I think it'd be best to merge now and get tests added in later, if that's fine with you.

And no problem! Looks like the main issue at hand was `dict` data objects being treated as iterables, so when no mapping was provided, then it'd just return a list of empty dicts for each key in the dictionary, e.g. `[{"key_1": "value_1"}, {"key_2": "value_2"}, {"key_3": "value_3"}]` would return `[{}, {}, {}]`

django_elasticsearch_dsl/fields.py
```diff
@@ -125,7 +128,7 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
         if objs is None:
             return {}
         try:
-            is_iterable = bool(iter(objs))
+            is_iterable = bool(iter(objs)) and not isinstance(objs, dict)
```
safwanrahman, line 131: I think rather than changing `is_iterable`, change to `if is_iterable and not isinstance(objs, dict):` below.

django_elasticsearch_dsl/fields.py
```diff
@@ -125,7 +128,7 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
...
         except TypeError:
             is_iterable = False
```
safwanrahman, line 134: Like this.
```suggestion
            if is_iterable and not isinstance(objs, dict)::
                ....
```

safwanrahman:
I am concerned that if anyone use `JsonField` in DB and add `ObjectField` with `properties`, will it filter out the other values?

oehrlein:
@safwanrahman Good call on moving the iterable dict check, it makes more sense with what you suggested. I committed the change.

oehrlein:
> I am concerned that if anyone use `JsonField` in DB and add `ObjectField` with `properties`, will it filter out the other values?

@safwanrahman If there are `properties` in the `ObjectField` and a field isn't defined, it will not show up. Although, if no `properties` are defined, then all fields will show up. Similarly, if there is an `ObjectField` in a `NestedField` (which is my personal use case), it works the same.

For example, using an `ObjectField` and this source data:
```
# Source data
{"make": "BMW", "model": "M5"}
```

Without defined properties:
```
# documents.py
class Car(Document):
    data = fields.ObjectField()

# Index
"_source" : {
    "data": {
        "make": "BMW",
        "model": "M5"
    }
}
```

With partially defined properties:
```
# documents.py
class Car(Document):
    data = fields.ObjectField(
        properties={
            'make': fields.TextField()
        }
    )

# Index
"_source" : {
    "data": {
        "make": "BMW"
    }
}
```

With all defined properties:
```
# documents.py
class Car(Document):
    data = fields.ObjectField(
        properties={
            'make': fields.TextField(),
            'model': fields.TextField()
        }
    )

# Index
"_source" : {
    "data": {
        "make": "BMW",
        "model": "M5"
    }
}
```

For example, using a `NestedField` and this source data:
```
# Source data
[{"make": "BMW", "model": "M3"}, {"make": "BMW", "model": "M5"}]
```

Without defined properties:
```
# documents.py
class Car(Document):
    data = fields.NestedField()

# Index
"_source" : {
    "data": [
        {
            "make": "BMW",
            "model": "M3"
        },
        {
            "make": "BMW",
            "model": "M5"
        }
    ]
}
```

With partially defined properties:
```
# documents.py
class Car(Document):
    data = fields.NestedField(
        properties={
            'make': fields.TextField()
        }
    )

# Index
"_source" : {
    "data": [
        {
            "make": "BMW"
        },
        {
            "make": "BMW"
        }
    ]
}
```

With all defined properties:
```
# documents.py
class Car(Document):
    data = fields.NestedField(
        properties={
            'make': fields.TextField(),
            'model': fields.TextField()
        }
    )

# Index
"_source" : {
    "data": [
        {
            "make": "BMW",
            "model": "M3"
        },
        {
            "make": "BMW",
            "model": "M5"
        }
    ]
}
```

oehrlein:
@safwanrahman Following up on tests, I was able to add some. See commit 31c8857.

safwanrahman:
@oehrlein That looks excellent! 🤩🚀
I will test it locally before merging it. Thanks a lot for adding the tests!

django_elasticsearch_dsl/fields.py
```diff
@@ -128,8 +131,8 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
             is_iterable = bool(iter(objs))
         except TypeError:
             is_iterable = False
-        
-        if is_iterable:
+
```
safwanrahman, line 134: Can you add a comment here why we are excluding the dictionary?
Something like
```suggestion
# Need to exclude dictionary, as it is iterable, but the full data need to be indexed.
```

oehrlein:
@safwanrahman No problem!

I also just added in comments for the two changes!

oehrlein:
@safwanrahman just wanted to follow up on this - is there anything else you need or can it be merged?

safwanrahman:
@oehrlein I am so sorry that I could merge it earlier! Was quite busy dealing with life things! 😓

safwanrahman:
@oehrlein I have rebased and merged it to master. Thanks a lot! 👍 💯

oehrlein:
@safwanrahman No worries, I appreciate you taking the time to do this!

PR #303, opened 2020-10-31
[Fix #296] Adding SearchAsYouType Field

### LINKED ISSUE #296: SearchAsYouType Field
Elastic Search DSL just added support for the Search As You Type field. Can we get this added to Django-Elasticsearch? I can do a PR if no one is doing this - I need this field in my production.

https://github.com/elastic/elasticsearch-dsl-py/commit/e881fc1506dcce6a1f07eea2c5d8a9956b3b925a

safwanrahman:
This fixes #296. It should be a small changes. As the field is added at es-dsl version 7.2.0, we should require that version in setup.py
@alexgarel r?

### DIFF
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -23,7 +23,8 @@ from elasticsearch_dsl.field import (
     Object,
     Short,
     Keyword,
-    Text
+    Text,
+    SearchAsYouType,
 )
 
 from .exceptions import VariableLookupError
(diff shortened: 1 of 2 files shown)


barseghyanartur:
Thumbs up for the `SearchAsYouType` addition. This, however, brings us to the branching hell. I still think that one codebase with tiny shims won't hurt at all. Since we can have ES version checks at no cost at all, we could define dummy classes and raise exceptions in case of version/feature conflicts.

safwanrahman:
@barseghyanartur I can understand that! But this is how the upstream package is maintaining their version and I think its easier for us to align with them!

PR #323, opened 2021-02-12
Improved support for refresh during updates

chosak:
This PR includes two commits that both improve the ability to refresh during updates.

First, it adds the ability to pass `refresh='wait_for'` to `Document.update()`. Currently you can either pass nothing (in which case no refresh happens) or `refresh=True`, in which case the refresh happens right away. [Per the docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/docs-refresh.html), it's also possible to specify `refresh='wait_for'` to

> [w]ait for the changes made by the request to be made visible by a refresh before replying. This doesn’t force an immediate refresh, rather, it waits for a refresh to happen.

Second, it adds a new `--refresh` parameter to the [search_index](https://django-elasticsearch-dsl.readthedocs.io/en/latest/management.html) management command. Currently when running that command, there's no way to force a refresh to happen. This can be useful, for example, if you want to programmatically call the command from some Django tests that fill an index, and then immediately run tests against that index. Without setting `refresh=True`, there's no guarantee that the refreshed index will be available to search.

See commit messages for additional details.

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -191,10 +191,10 @@ class DocType(DSLDocument):
         """
         Update each document in ES for a model, iterable of models or queryset
         """
-        if refresh is True or (
-            refresh is None and self.django.auto_refresh
-        ):
+        if refresh is None and self.django.auto_refresh:
             kwargs['refresh'] = True
+        elif refresh is not None:
+            kwargs['refresh'] = refresh
 
         if isinstance(thing, models.Model):
             object_list = [thing]
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -191,10 +191,10 @@ class DocType(DSLDocument):
         """
         Update each document in ES for a model, iterable of models or queryset
         """
-        if refresh is None and self.django.auto_refresh:
-            kwargs['refresh'] = True
-        elif refresh is not None:
+        if refresh is not None:
             kwargs['refresh'] = refresh
+        elif self.django.auto_refresh:
+            kwargs['refresh'] = True
 
         if isinstance(thing, models.Model):
             object_list = [thing]
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -194,7 +194,7 @@ class DocType(DSLDocument):
         if refresh is not None:
             kwargs['refresh'] = refresh
         elif self.django.auto_refresh:
-            kwargs['refresh'] = True
+            kwargs['refresh'] = self.django.auto_refresh
 
         if isinstance(thing, models.Model):
             object_list = [thing]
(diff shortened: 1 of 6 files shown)


safwanrahman:
@chosak Thanks a lot for the PR. I think it was better to create a issue before submitting a patch. So we could have discussed how this could be implemented!

As you have submitted the patch, I will review it and will surely try to merge it to master. Thanks

django_elasticsearch_dsl/documents.py
```diff
@@ -191,10 +191,10 @@ def update(self, thing, refresh=None, action='index', parallel=False, **kwargs):
         """
         Update each document in ES for a model, iterable of models or queryset
         """
-        if refresh is True or (
-            refresh is None and self.django.auto_refresh
-        ):
+        if refresh is None and self.django.auto_refresh:
             kwargs['refresh'] = True
+        elif refresh is not None:
+            kwargs['refresh'] = refresh
```
safwanrahman, line 197: I think if `refresh` parameter is passed, it should be provided the highest priority regardless what is mentioned in the document. What do you think? @chosak

tests/test_documents.py
```diff
@@ -268,6 +268,26 @@ def test_model_instance_update_no_refresh(self):
...
+        doc = CarDocument()
+        doc.django.auto_refresh = False
+        car = Car()
+        with patch('django_elasticsearch_dsl.documents.bulk') as mock:
+            doc.update(car, refresh='wait_for')
+            self.assertEqual(
+                mock.call_args_list[0][1]['refresh'], 'wait_for'
+            )
+
+    def test_model_instance_update_refresh_wait_for_overrides_auto_refresh(self):
```
safwanrahman, line 281: Can you add a test with `refresh=true` as well?

safwanrahman:
@chosak I have couple of simple issue about this PR, other than that it looks really good. Can you fix that so we can get it merged?

chosak:
:+1: @safwanrahman thanks for the review!

django_elasticsearch_dsl/documents.py
```diff
@@ -191,9 +191,9 @@ def update(self, thing, refresh=None, action='index', parallel=False, **kwargs):
         """
         Update each document in ES for a model, iterable of models or queryset
         """
-        if refresh is True or (
-            refresh is None and self.django.auto_refresh
-        ):
+        if refresh is not None:
+            kwargs['refresh'] = refresh
+        elif self.django.auto_refresh:
```
safwanrahman, line 196: @chosak I think people should have the ability to set the `wait_for` in the document as well. What about setting like following?
```suggestion
        elif self.django.auto_refresh:
            kwargs['refresh'] = self.django.auto_refresh
 ```
```

chosak: Makes sense. I've added a test for this and updated the documentation as well. The next step would be to allow `settings.ELASTICSEARCH_DSL_AUTO_REFRESH` to support `wait_for` as well...

safwanrahman: Sounds interesting! I need to think the naming as well. Because `auto_refresh` does not make sense anymore. But we can think about it in future. 💯

Issue #290, opened 2020-08-03
Allow rebuilding specific index.


jbonet:
Hello,

I find myself having two indexes for a model. While we transition to the new one, we need to keep the old one. But rebuilding in development won't ask for which one I want to rebuild, it just recreates both.

It would be nice if it asked individually for each index in a model.

rafascar:
Hey! I also found myself in need of this feature in multiple occasions.

Is anyone working on this feature? If not, I would be interested to contribute, and gladly accept any opinions on how to approach this.

Cheers!

PR #328, opened 2021-03-06
Add --indices option to search_index management command

rafascar:
Add a new `--indices` option to the `search_index` Django management command, that allows issuing commands (i.e. `--create`, `--populate`, `--delete` and `--rebuild`) for specific indices referenced by their names.

The current implementation only allows referencing indices by their models. This doesn't allow e.g. rebuilding a specific index for a model that has multiple indices, which may be useful when transitioning from an old index to a new index with different and possibly breaking mappings.

Issue #290


safwanrahman:
As I mentioned earlier in #237, I am not sure about this feature because [index names are not unique](https://github.com/django-es/django-elasticsearch-dsl/pull/237#issuecomment-586266182), so it may leads to uncertain behavior.

rafascar:
> As I mentioned earlier in #237, I am not sure about this feature because index names are not unique, so it may leads to uncertain behavior.

Aw, that's sad, completely missed this one 😢 thanks for the reply though, I'll close the PR until we can figure something out. 👋

PR #329, opened 2021-03-07
Fix minor documentation typos

chosak:
- Use of Markdown link instead of reStructuredText.
- Typo: wan't instead of want


safwanrahman:
Thanks!

PR #343, opened 2021-04-27
Set default_app_config only in Django 3.1 or less

alexgmin:
https://docs.djangoproject.com/en/3.2/releases/3.2/#automatic-appconfig-discovery

default_app_config is deprecated starting in 3.2, so a project in that version using this library with Deprecation Warnings enabled gets a warning about it.

This pull request sets the variable only if the django version is previous to 3.2


safwanrahman:
LGTM! r+
Thanks @alexgmin

PR #346, opened 2021-05-14
Allow for ObjectFields to call prepare_func included in InnerDoc.

paulogiacomelli:
These changes allow for adding prepare_functions inside subclassed `InnerDoc`. This is very useful if documents need to index some Django model which needs to include some prepare_fields.

For example, we can have a `ManufacturerInnerDoc` that’s included as a `manufacturer = fields.ObjectField(doc_class=ManufacturerInnerDoc)`, and also as class `ManufacturerDocument(Document, ManufacturerInnerDoc)`.

This allows for more reusability of fields in different indexes and on its own document. We can now only defined 1 `InnerDoc` and use on the main Document (if needed) and any other Document.

### DIFF
--- a/django_elasticsearch_dsl/fields.py
+++ b/django_elasticsearch_dsl/fields.py
@@ -114,10 +114,10 @@ class ObjectField(DEDField, Object):
 
                 # This allows for retrieving data from an InnerDoc with prepare_field_name functions.
                 doc_instance = self._doc_class()
-                fn = getattr(doc_instance, 'prepare_%s' % name, None)
+                prep_func = getattr(doc_instance, 'prepare_%s' % name, None)
 
-                if fn:
-                    data[name] = fn(obj)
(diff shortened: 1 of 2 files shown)


safwanrahman:
Thanks for the Pull request, can you create a new issue with the problem or bug you were facing. I am understanding you code, but could not find any way to test it.

saadmk11 (commented):
Thanks for the pull request! :rocket:  I was looking for this feature, Changes look awesome. :100: 
Added some small nitpicks with the review. but nothing important.

Running this through the whole test suite would be great but the CI is not running it seems.

paulogiacomelli:
@saadmk11 Thanks for the review. I have updated the PR.

safwanrahman:
LGTM. r+ Thanks a lot 💯  @paulogiacomelli

PR #349, opened 2021-06-04
Add `post_index` signal

saadmk11:
Closes https://github.com/django-es/django-elasticsearch-dsl/issues/298

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -149,6 +149,7 @@ class DocType(DSLDocument):
         post_index.send(
             sender=self.__class__,
             instance=self,
+            actions=actions,
             response=response
         )
         return response
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -23,6 +23,7 @@ from .fields import (
     TextField,
 )
 from .search import Search
+from .signals import post_index
 
 model_field_class_to_field_class = {
     models.AutoField: IntegerField,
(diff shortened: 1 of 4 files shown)


safwanrahman:
Thanks @saadmk11 for the PR. It looks very nice with the docuemntation. I will take a closer look into it.

django_elasticsearch_dsl/documents.py
```diff
@@ -143,7 +144,14 @@ def to_field(cls, field_name, model_field):
             )
 
     def bulk(self, actions, **kwargs):
-        return bulk(client=self._get_connection(), actions=actions, **kwargs)
+        response = bulk(client=self._get_connection(), actions=actions, **kwargs)
+        # send post index signal
+        post_index.send(
+            sender=self.__class__,
+            instance=self,
```
safwanrahman, line 151: I think you should pass the actions as well to the signal. Without it, it is quite hard to determine which object is indexed to Elasticsaerch.

saadmk11:
@safwanrahman Thanks for reviewing the PR, I have updated the PR with the suggested changes.

tests/test_signals.py
```diff
@@ -0,0 +1,41 @@
...
+        car = Car(
+            pk=51,
+            name="Type 57"
+        )
+
+        doc.update(car)
+
+        bulk.assert_called_once()
+
+        post_index.send.assert_called_once_with(
```
safwanrahman, line 36: It is better to register a signal function and check whether the function is called with proper values

saadmk11: @safwanrahman Updated the tests.

safwanrahman:
Looks Good to me! r+ 🤩🚀
@saadmk11 you rock! 🥳

PR #355, opened 2021-06-29
Add option to update only specific fields while indexing documents

saadmk11:
This PR adds an option to update only specific fields of a document in Elasticsearch. See the referenced issue for more details.

Some things to think about:

On `Model.save(update_fields=['some_field'])`, the `update_fields=['some_field']` is also passed to the signals and the signals pass that to `registry.update()` method. Though it can be a good thing but can create issues if not carefully used. such as not updating non-model fields on the document that depends on the updated fields as only `'some_field'` will be updated.

These changes also require some documentation as well, but I'm not sure where to put it. We can put the docs on the `Index` documentation page.

Closes https://github.com/django-es/django-elasticsearch-dsl/issues/277

CC: @safwanrahman

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -119,15 +119,20 @@ class DocType(DSLDocument):
         """
         Take a model instance, and turn it into a dict that can be serialized
         based on the fields defined on this DocType subclass
+        optionally, update_fields can be passed to index only specific fields
         """
-        data = {}
-
-        for name, field, prep_func in self._prepared_fields:
-            if isinstance(update_fields, (list, tuple, set)):
-                if name in update_fields:
-                    data.update({name: prep_func(instance)})
-            else:
-                data.update({name: prep_func(instance)})
+        if isinstance(update_fields, (list, tuple, set)):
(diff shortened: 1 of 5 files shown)


safwanrahman:
I am not sure it is a feature that is needed because we can not map the dependencies of a field. Like `Should field B should be updated when A is also updated?`

safwanrahman:
If this feature need to be merged, there should be some configuration to activate/deactivate it. And by default it should be deactivated.

saadmk11:
> I am not sure it is a feature that is needed because we can not map the dependencies of a field. Like `Should field B should be updated when A is also updated?`

Yes, I agree with you, Even if we put this behind a setting it is bound to cause confusion among users and cause weird edge cases if not used carefully. We may choose to not implement this at all.

@safwanrahman feel free to close this if you agree. otherwise, I will add a setting for this.

safwanrahman:
@saadmk11 I agree with the concern. Thanks a lot the PR, but it is better if we can close it. 😅

konman1989:
@saadmk11 @safwanrahman do you have any plans to implement such a feature? Or a workaround for this issue?

Issue #314, opened 2020-12-20
Rebuild index with one down time


jphilip:
Hi, thanks for all the work.
I was trying to write a task to rebuild an index without down time on a single node setup and based on [this ](https://gist.github.com/rlam3/8781648c7f311de41fa8)example. Basically it creates a new index and swaps aliases to replace the existing index because an index cannot be renamed. I could make it work, but my new alias is not being used by django-elasticsearch-dsl because it seems to only rely on the index name, not an alias. It takes only 1.2 minutes to rebuild my index at this time, so the down time is not too bad, but is there any way to do this currently, or in the future? It would be nice to be able to rebuild indexes without down time.

jhofeditz:
Instead of an alias, I solved this by adding a second document subclassing the original one so I can use the dsl in my scripts. Then I have a script like the one you linked to that rebuilds the tmp index and then copies from the tmp index to the public one without downtime.
```@registry.register_document
class ModelDocument(ModelTmpDocument):
    """The actual user facing index"""
    class Index:
        name = es_index

**Comment by antunesleo** [comment]
I think this rebuild strategy can lead to information loss, what happens with documents being created while the rebuilding is running?

I'm currently studying a way to do implement rebuilding with no downtime, I've found this strategy interesting: https://medium.com/craftsmenltd/rebuild-elasticsearch-index-without-downtime-168363829ea4

Despite interesting, it's complex to implement since we rely in two different aliases, one for **reads** and another one for **writes**. 
While the migration is being executed, our models running in production code must be able to insert data in the correct index, since we don't want to loss any data created while the migration is running.

To implement the strategy, I think we could do a _"CQRS Style"_, using 2 models:  one for reads and another one for writes, e.g

```
class CQRSDocument:
    pass

class ExampleRead(CQRSDocument):
    """The actual user facing index"""
    class Index:
        name = read_example

class ExampleWrite(CQRSDocument):
    """The actual user facing index"""
    class Index:
        name = write_example
```

**read_example** and **write_example** are aliases, so we can change aliases index's while the migration is running as suggested by the article. That way we can rebuild it with no downtime, with no data loss and without changing application's code.

We also would need to refactor exiting commands **build** and **rebuild** to do the new rebuilding strategy when the model class is **CQRSDocument** instead of **Document**.

[...]

PR #358, opened 2021-07-09
Add atomic/zero-downtime index rebuilding

oehrlein:
These changes allow for atomic/zero-downtime index rebuilding using aliases.

There are two new management commands: `--use-alias` and `--use-alias-keep-index`, both which only work with the `--rebuild` command. If `--use-alias-keep-index` is used with `--use-alias`, the indices previously aliased will not be deleted. If `--use-alias-keep-index` isn't used, the old indices will be deleted.

Regarding naming, the defined index name in `documents.py` becomes the alias name, and indices take the alias name plus a suffix (hyphen + current datetime). I took example from [this](https://github.com/elastic/elasticsearch-dsl-py/blob/master/examples/alias_migration.py).

2 big caveats:

1. This only works in the simplest use case of just rebuilding and replacing indices - it won't work in the event of continuous updates to an index, as outlined in [this article](https://medium.com/craftsmenltd/rebuild-elasticsearch-index-without-downtime-168363829ea4) in the section "Wait, I know a case when this will not work…"
2. As mentioned in [this example](https://github.com/elastic/elasticsearch-dsl-py/blob/master/examples/alias_migration.py) (same example as above), in order to properly deserialize the data in searches, you will need to add this function to your documents in `documents.py`:
```
@classmethod
def _matches(cls, hit):
    # override _matches to match indices in a pattern instead of just ALIAS
    # hit is the raw dict as returned by elasticsearch
    return fnmatch(hit["_index"], ALIAS + "-*")
```
For this to work, remember to add `from fnmatch import fnmatch` to your imports in `documents.py`. Also, replace `ALIAS` with your alias name, or just make that argument one string, e.g. `'my-alias-*'`.

As noted in the example:
> BlogPost._matches() class method is required for this code to work since
      otherwise BlogPost will not be used to deserialize the documents as those
      will have index set to the concrete index whereas the class refers to the
      alias. 

I didn't get around to writing tests or adding to the documentation. With all that being said, please let me know if you have questions and/or suggestions!

**EDIT: Please note that caveat 2 has been taken care of in ce05715, so it is no longer an issue.**


oehrlein:
This resolves #314

safwanrahman:
Thanks for the Pull request, I will review it and add my concern into it. @saadmk11 Can you take a look as well?

saadmk11:
Looks like there is a similar PR https://github.com/django-es/django-elasticsearch-dsl/pull/284 that does this as a part of it but has been stale for a long time

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -63,6 +65,18 @@ def add_arguments(self, parser):
             dest='parallel',
             help='Run populate/rebuild update single threaded'
         )
+        parser.add_argument(
+            '--atomic',
```
saadmk11, line 69: The naming of the flags look a little bit confusing to me, maybe we could use something like `--use-alias`, what do you think?

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -139,13 +153,85 @@ def _delete(self, models, options):
             index.delete(ignore=404)
         return True
 
+    def _update_alias(self, es_conn, alias, new_index, alias_exists, options):
+        alias_actions = [{"add": {"alias": alias, "index": new_index}}]
+        old_indices = []
+        alias_delete_actions = []
+        if alias_exists is True:
```
saadmk11, line 160:
```suggestion
        if alias_exists:
```

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -139,13 +153,85 @@ def _delete(self, models, options):
...
+            else:
+                stdout_term = "indices"
+
+            old_indices_str = ", ".join(old_indices)
+            self.stdout.write(
+                "Removed alias '{}' from {} '{}'".format(
+                    alias, stdout_term, old_indices_str
+                )
+            )
+            if alias_delete_actions and options['atomic_no_delete'] is False:
```
saadmk11, line 190:
```suggestion
            if alias_delete_actions and not options['atomic_no_delete']:
```

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -139,13 +153,85 @@ def _delete(self, models, options):
...
+                es_conn.indices.update_aliases(
+                    {"actions": alias_delete_actions}
+                )
+                self.stdout.write(
+                    "Deleted {} '{}'".format(stdout_term, old_indices_str)
+                )
+
     def _rebuild(self, models, options):
-        if not self._delete(models, options):
+        if options['atomic'] is False and not self._delete(models, options):
```
saadmk11, line 199:
```suggestion
        if not options['atomic'] and not self._delete(models, options):
```

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -139,13 +153,85 @@ def _delete(self, models, options):
...
             return
 
+        if options['atomic'] is True:
```
saadmk11, line 202:
```suggestion
        if options['atomic']:
```

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -139,13 +153,85 @@ def _delete(self, models, options):
...
+                new_index = index._name + index_suffix
+                alias_index_pairs.append(
+                    {'alias': index._name, 'index': new_index}
+                )
+                index._name = new_index
+
         self._create(models, options)
         self._populate(models, options)
 
+        if options['atomic'] is True:
```
saadmk11, line 218:
```suggestion
        if options['atomic']:
```

saadmk11 (commented):
Thank you for the pull request :100: , added some small suggestions.

@safwanrahman I have not tested this locally yet will do it when I get some time.

oehrlein:
> Looks like there is a similar PR #284 that does this as a part of it but has been stale for a long time

I tried out the changes in that PR, and unless I misunderstood what was documented, it didn't do exactly what I was looking for. Regardless, yea, it's been stale for a while.

django_elasticsearch_dsl/management/commands/search_index.py
oehrlein: `--use-alias` works for me - overall my naming on the flags was something I wasn't to set on. Similarly, I think `--atomic-no-delete` could/should now be `--use-alias-no-delete-indices`.

At the moment, the default behavior of `--use-alias` is to delete the old indices. Do you think that is fine in being the default, or should the deletion of the old indices have to be explicit via another flag? I.e. get rid of `--use-alias-no-delete-indices` and add something like `--use-alias-delete-indices`.

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -139,13 +153,85 @@ def _delete(self, models, options):
...
             return
 
+        if options['atomic']:
+            alias_index_pairs = []
+            index_suffix = "-" + datetime.now().strftime("%Y%m%d%H%M%S%f")
+            for index in registry.get_indices(models):
+                # The alias takes the original index name value. The
+                # index name sent to Elasticsearch will be the alias
+                # plus the suffix from above.
+                new_index = index._name + index_suffix
```
safwanrahman, line 209: There are limit for [index name](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-path-params) 255 charector. Can you limit slice the string so it never exceeds 255 charector?

safwanrahman:
This seems like a good approach. Though I was thinking to different way to achieve it, but it will achieve the purpose with minimal changes. I will test it locally before mege. I have added a little fix over.

antunesleo:
First of all, thanks for tackling this one @oehrlein 

About the  **continuous updates to an index** issue, I think we can mitigate it running combined your solution with populate command, this way:

First we run **search_index --rebuild --atomic**
secondly we run a  **search_index populate** in order to index documents that were inserted during the migration

One caveat is that items inserted during the migration will be inconsistent for a brief moment, the items will not be available for search until we ran the populate command. 
I believe this is acceptable for search engines like ES that **are not strongly consistent** and the most important thing: we **are not loosing any document** during the migration.

django_elasticsearch_dsl/management/commands/search_index.py
oehrlein: Yes, just committed the changes!

oehrlein:
@safwanrahman Sounds good, I appreciate it!

django_elasticsearch_dsl/management/commands/search_index.py
oehrlein: @safwanrahman What are your thoughts on this?

oehrlein: @saadmk11 @safwanrahman Just wanted to follow up - what are your thoughts?

safwanrahman: @oehrlein Sorry for not getting back to it. I was quite busy with some issues. I think `--use-alias` is better suited for the term. I think the deletion should be by default, we can add another flag that will not delete the old index. What do you think?

oehrlein: @safwanrahman No worries, I was busy over the past week as well! 

I agree that `--use-alias` is the best choice and went ahead and made the changes. Likewise, I kept deletion as the default behavior and renamed `--atomic-no-delete` to `--use-alias-keep-index`.

oehrlein:
@safwanrahman @saadmk11 I made the changes for all your guys' suggestions, so it should all be good to merge!

safwanrahman:
Thanks a lot @oehrlein  for your patience  and patch that is very good design. @saadmk11 Can you test it with any application and check if it is working perfectly? I am being too busy with personal things so maybe it would be better if you can test it and let me know. I will merge it after you test it.

oehrlein:
@safwanrahman No problem, and thanks!

@saadmk11 Let me know if anything needs to be changed!

saadmk11:
> @saadmk11 Let me know if anything needs to be changed!

Hi @oehrlein, Thanks for updating the PR with the requested changes. :100: 

I tried to test these changes locally (by running `python manage.py search_index --rebuild --use-alias`) on an application that already had an index created on the Elasticsearch cluster (using the rebuild command of the current version of `django-elasticsearch-dsl`).

I am getting this error when I run the command:

```console
Traceback (most recent call last):
  File "manage.py", line 21, in <module>
    main()
  File "manage.py", line 17, in main
    execute_from_command_line(sys.argv)
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 262, in handle
    self._rebuild(models, options)
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 242, in _rebuild
    options
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/django_elasticsearch_dsl/management/commands/search_index.py", line 179, in _update_alias
    es_conn.indices.update_aliases({"actions": alias_actions})
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/elasticsearch/client/utils.py", line 301, in _wrapped
    return func(*args, params=params, headers=headers, **kwargs)
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/elasticsearch/client/indices.py", line 612, in update_aliases
    "POST", "/_aliases", params=params, headers=headers, body=body
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/elasticsearch/transport.py", line 458, in perform_request
    raise e
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/elasticsearch/transport.py", line 426, in perform_request
    timeout=timeout,
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/elasticsearch/connection/http_urllib3.py", line 287, in perform_request
    self._raise_error(response.status, raw_data)
  File "/home/saad/.virtualenvs/test_project/lib/python3.6/site-packages/elasticsearch/connection/base.py", line 338, in _raise_error
    status_code, error_message, additional_info
elasticsearch.exceptions.RequestError: RequestError(400, 'invalid_alias_name_exception', 'Invalid alias name [some_index_name], an index exists with the same name as the alias')
```
(I think this is happening because we are trying to create an alias with the same name of the index. As the index name does not have any timestamp/uniqueness on the current implementation it is conflicting with the alias name)

But If I delete the index on ElasticSearch and then run the rebuild command it works as expected because the newly created index name has timestamp/uniqueness according to the PR implementation.

Please let me know If I need to add/update anything to test the changes of this PR. :)

oehrlein:
@saadmk11 No problem!

Thanks for catching this error! I ran into it before, but forgot about it before my original commits.

I made adjustments to fix the issue. Now, when we want an alias to have the name of an index that already exists, it deletes the existing index as well. Likewise, I made changes so that when we want an index to have the name of an alias that already exists, it deletes the indices pointed at the alias to free up the name.

I also made adjustments to account for deleting indices that were created with aliases (after running `--rebuild` with `--use-alias`, run `--delete` with `--use-alias` to delete those indices). 

In addition, I added additional stdout messages in case of doing something alias-specific, but only having regular indices, or vice-versa. For example, trying to delete regular indices after creating indices with an alias (after running `--rebuild` with `--use-alias`, run `--delete` without `--use-alias`) provides a message explaining the situation and how to resolve it.

Take a look at the changes when you get a chance, and let me know if you have any suggestions and/or think there should be changes!

safwanrahman:
@oehrlein Thanks a lot for updating. The test are failing unfortunately. 😢 Can you fix it?

oehrlein:
@safwanrahman I fixed the issues from my end that were causing the tests to fail, but looks like tests are failing for everyone due to another issue (#377).

oehrlein:
@safwanrahman I was able to also fix the test that was failing for everyone! All the checks have passed :)

safwanrahman:
@oehrlein Thanks a lot for fixing the tests and other fixes. As the test is failing, I have cherry-picked your test fix commit and pushed that commit only to master. I will merge this PR after @saadmk11 give it another try.
@saadmk11 Will you have some time to review the PR code and test it locally please? It would be much appreciated.

oehrlein:
@safwanrahman Thanks, and sounds good!

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -123,29 +152,139 @@ def _populate(self, models, options):
...
+                new_index = index._name[:234] + index_suffix
+                alias_index_pairs.append(
+                    {'alias': index._name, 'index': new_index}
+                )
+                index._name = new_index
+        else:
+            for index in registry.get_indices(models):
+                alias_exists = index._name in aliases
+                if alias_exists:
+                    self._delete_alias_indices(index._name)
```
saadmk11, line 275: is it possible to move this code block to the `_delete()` method?

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -156,14 +295,18 @@ def handle(self, *args, **options):
         action = options['action']
         models = self._get_models(options['models'])
 
+        aliases = []
+        for index in self.es_conn.indices.get_alias().values():
+            aliases += index['aliases'].keys()
```
saadmk11, line 300: Can we only do this if `--use-alias` is True?

saadmk11 (commented):
@oehrlein Great work on this! :100: 

The management commands are working pretty well for me. :)

Regarding the issue with `_matches` we can override it on the `DocType` class of the `documents.py` file. It may return something similar to this:

```python
return fnmatch(hit.get("_index", ""), f'{cls._index._name}*')
```
This will cover both alias and non-alias use cases. One concern here is if multiple indices are named similarly (e.g: `test_index`, `test_index2`) this will not be able to figure out which one to match.

Another option is to document this issue and tell the user to override this on their end when using aliases.

I have added some small concerns, but nothing major. Again Great work on this. :+1:

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -123,29 +152,139 @@ def _populate(self, models, options):
...
+
+        self.es_conn.indices.update_aliases({"actions": alias_actions})
+        if delete_existing_index:
+             self.stdout.write("Deleted index '{}'".format(alias))
+
+        self.stdout.write(
+            "Added alias '{}' to index '{}'".format(alias, new_index)
+        )
+
+        if old_indices:
```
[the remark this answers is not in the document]
oehrlein: I have this in case of `old_indices` being empty (e.g. on the first run of the script, so no old indices exist). Let me know if that makes sense.

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -123,29 +152,139 @@ def _populate(self, models, options):
...
+    def _update_alias(self, alias, new_index, alias_exists, options):
+        alias_actions = [{"add": {"alias": alias, "index": new_index}}]
+
+        delete_existing_index = False
+        if not alias_exists and self.es_conn.indices.exists(index=alias):
+            # Elasticsearch will return an error if an index already
+            # exists with the desired alias name. Therefore, we need to
+            # delete that index.
+            delete_existing_index = True
+            alias_actions.append({"remove_index": {"index": alias}})
```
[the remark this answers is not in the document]
oehrlein: By warning, do you just mean a stdout message explaining why the index is being deleted? If so, that's fine with me.

django_elasticsearch_dsl/management/commands/search_index.py
oehrlein: Good call on this one! I updated it in 5b53d5a.

django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -123,29 +152,139 @@ def _populate(self, models, options):
...
+                alias_exists = index._name in aliases
+                if not alias_exists:
+                    self.stdout.write("Deleting index '{}'".format(index._name))
+                    index.delete(ignore=404)
+                elif options['action'] == 'delete':
+                    self.stdout.write(
+                        "'{}' refers to an alias, not an index. Run "
+                        "'--delete' with '--use-alias' arg to delete indices "
+                        "pointed at the alias.".format(index._name)
+                    )
```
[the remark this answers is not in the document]
oehrlein: Also as above, I updated it in 39d1ba2!

oehrlein:
@saadmk11 Thanks for the kind words and the help + suggestions!

I took your suggestion regarding the `_matches` function and added it in ce05715. While I understand that having multiple, similarly-named indices will cause matching issues, I think that's less likely to happen than someone trying to use aliases and forgetting to override the `_matches` function on their own.

There's just 2 suggestions that I'm looking for your response for - everything else should be good! :)

saadmk11:
> I took your suggestion regarding the `_matches` function and added it in [ce05715](https://github.com/django-es/django-elasticsearch-dsl/commit/ce057156354ba63b823cc5eb9fb01dd0125704a5). While I understand that having multiple, similarly-named indices will cause matching issues, I think that's less likely to happen than someone trying to use aliases and forgetting to override the `_matches` function on their own.

Thanks for adding this. I would like to know if @safwanrahman has any thoughts on this.

oehrlein:
@safwanrahman Any luck at being able to take a look at this?

oehrlein:
@safwanrahman Just wanted to follow up again - any luck at being able to take a look at this?

violuke:
We're really like to see this too @safwanrahman if you are able to review. Thank you for your time

rsommerard:
Hi, any news about the merge of this **really useful** feature?

Root-kjh:
i found that this source code does not erase new indexes when an error occurs in process of creating alias

oehrlein:
@phpaulin I definitely would like to, but it's beyond me unfortunately.

@safwanrahman @saadmk11 Any chance you guys will have time soon to look at this again and merge it?

phpaulin:
Yes, it would be greatly helpful for our team also @safwanrahman .

safwanrahman:
@oehrlein Thanks a ton for the PR. I have been quite busy for some months so could not verify to merge it. Finally, I got some time to check the code and it seems to be good from my end.

I have merged it now! 💯 Thanks again for your patience.

violuke:
Incredible, thanks so much @safwanrahman

oehrlein:
@safwanrahman No problem, and thanks for taking a look at it and merging it!

@violuke No problem!

zillasaur:
@safwanrahman will the release be bumped to include this soon? i'd love to use this functionality!

zillasaur:
@oehrlein thank you for doing this, amazing work! is it possible to update the PR description with the correct management command names and also have the proper documentation for this new feature? (pending on the new version)

fizikurniawan:
@oehrlein Thank you for the PR. I have a lot of data to index, what will happen if there is new data during the indexing process? Will that data be stored in the new index (when use --use-alias) or not indexed? Because average time to populate takes about 20 min.

oehrlein:
@zillasaur Firstly, my apologies for the long delay to get back to you.

I updated the PR description with the correct management commands. I also added documentation for the management commands and submitted a PR (#446) for it.

oehrlein:
@fizikurniawan Firstly, my apologies for the long delay to get back to you.

To answer your question, I'm not sure. Ultimately, that's more of a question of how populating an Elasticsearch index works, considering populating an index with or without aliases works the same.

PR #363, opened 2021-07-25
Move the test suite to GitHub Actions

saadmk11:
This Pull Request moves the test suite to GitHub Actions from travis-ci.

Note:
codecov and travis-ci were not running previously.


saadmk11:
@safwanrahman Would be great to get your review here.

saadmk11:
Sample GitHub Action run: https://github.com/saadmk11/django-elasticsearch-dsl/actions/runs/1064789205

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+            django-version: "2.0"
+          - python-version: "3.8"
+            django-version: "2.1"
+
+    steps:
+      - name: Configure Sysctl Limits for Elasticsearch
+        run: |
+          sudo swapoff -a
+          sudo sysctl -w vm.swappiness=1
+          sudo sysctl -w fs.file-max=262144
```
safwanrahman, line 43: Do we really need swap?

safwanrahman: I believe 7GB ram would be enough to run the test suite!
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

tests/test_fields.py
```diff
@@ -50,7 +50,7 @@ def test_get_value_from_instance_with_unknown_attr(self):
         class Dummy:
             attr1 = "foo"
 
-        field = DEDField(attr='attr2')
+        field = DEDField(attr='attr2', required=True)
```
safwanrahman, line 53: Any particular reason to add it to required ?

.github/workflows/ci.yml
saadmk11: This is recommended by elastic:

> You must also add the Configure sysctl limits step, otherwise Elasticsearch will not be able to boot.

Ref: https://github.com/elastic/elastic-github-actions/tree/master/elasticsearch#usage

tests/test_fields.py
saadmk11: The test tries to assert `VariableLookupError` is raised. but `VariableLookupError` is not raised if required is not True.
This test also fails locally.

https://github.com/django-es/django-elasticsearch-dsl/blob/759076bb8e968db311610226966d448767e4441e/django_elasticsearch_dsl/fields.py#L71-L75

.github/workflows/ci.yml
saadmk11: @safwanrahman I have removed this and it looks like everything is working as expected, not sure why it was recommended by elastic, maybe there was some edge case before as the README was last updated more than a year ago.

Output: https://github.com/saadmk11/django-elasticsearch-dsl/actions/runs/1119509026

safwanrahman:
Looks very nice! Thanks @saadmk11

PR #370, opened 2021-09-03
Fix DocType._get_actions preventing delete

seljin:
Hi everybody, thank you for the good work here.

There is a case when ES index is not updated on instance deletion :

If we base our should_index_object method on manytomany field, when we delete an object, the related objects are deleted before the object itself and doesnt exists anymore when should_index_object method is called.

Thus, should_index_object return is incoherent in this case and the ES index is not updated.

I think we should not call should_index on delete action, what do you think ?

I also corrected some flake8 errors.

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -51,8 +51,10 @@ model_field_class_to_field_class = {
     models.UUIDField: KeywordField,
 }
 
+
 class DocType(DSLDocument):
     _prepared_fields = []
+
     def __init__(self, related_instance_to_ignore=None, **kwargs):
         super(DocType, self).__init__(**kwargs)
         self._related_instance_to_ignore = related_instance_to_ignore
@@ -125,8 +127,8 @@ class DocType(DSLDocument):
         """
         data = {
             name: prep_func(instance)
-                for name, field, prep_func in self._prepared_fields
-            }
+            for name, field, prep_func in self._prepared_fields
+        }
         return data
 
     @classmethod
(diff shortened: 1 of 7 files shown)


safwanrahman:
Interesting find. Can you let me know how to reproduce the issue? I was the person who implemented `should_index_object` API and it might be a good learning for me.

seljin:
Of course 👍 you can use this example : https://github.com/seljin/ded-should-index-example

I stay available if you need anything :)

saadmk11:
Great Find @seljin!  :100: 

Not calling `should_index_object()` for `delete` action should fix the issue with related objects.

But I am concerned about people who are already overriding  `should_index_object()` to customize delete action as well. So, not calling  `should_index_object()` for `delete` action will break their implementation.

Another way might be passing the `action` variable as a parameter to  `should_index_object()` so that the users can handle different actions differently but that will also break other people's implementation who are overriding  `should_index_object()`.

seljin:
Could you show me an example of how and why you customize `should_index_object()` for `delete` action please ?

I can't think of a case that you need to do that because when you delete an object in the database you need to delete it in the index no matter what. If you dont do that the index is not synced with your database.

Maybe there is something I cant see here, it might be a good learning for me if you could show me an example

safwanrahman:
@seljin Good point. What do you think @saadmk11 ?

saadmk11:
Yup, @seljin is right. I was thinking about something different which is not the case for this package. I am :+1:  with this.

seljin:
Great, thank you for your time @saadmk11 and @safwanrahman  👍

safwanrahman:
Looks Good! Thanks @seljin and @saadmk11

PR #378, opened 2021-10-03
fix log message in search_index populate command

denheck:
Hello and thanks for maintaining such an awesome library! 

I noticed while running the `search_index --rebuild` command that the number of documents being indexed is incorrect in the log message. It looks like this is because the count in the log message is filtered by `get_queryset` instead of the `get_indexing_queryset` which is used in the actual indexing process.

Thanks again!

### DIFF
--- a/django_elasticsearch_dsl/management/commands/search_index.py
+++ b/django_elasticsearch_dsl/management/commands/search_index.py
@@ -116,7 +116,7 @@ class Command(BaseCommand):
         parallel = options['parallel']
         for doc in registry.get_documents(models):
             self.stdout.write("Indexing {} '{}' objects {}".format(
-                doc().get_queryset().count() if options['count'] else "all",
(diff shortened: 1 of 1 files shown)


django_elasticsearch_dsl/management/commands/search_index.py
```diff
@@ -116,7 +116,7 @@ def _populate(self, models, options):
         parallel = options['parallel']
         for doc in registry.get_documents(models):
             self.stdout.write("Indexing {} '{}' objects {}".format(
-                doc().get_queryset().count() if options['count'] else "all",
+                doc().get_indexing_queryset().count() if options['count'] else "all",
```
saadmk11, line 119: As `get_indexing_queryset()` returns a `generator` object, I think we won't be able to invoke `count()` on it.

safwanrahman:
You should overwrite the `get_queryset` method if you do some filtering. `get_indexing_queryset` returns generator which will not work with `count`

PR #386, opened 2021-12-07
fields.py: force_text to force_str

HorebParraud:
https://docs.djangoproject.com/en/4.0/releases/4.0/#features-removed-in-4-0

without advancing too much I believe that it is the deleted import of Django 4


safwanrahman:
Already implemented by  #387

PR #387, opened 2021-12-07
Add Django 4.0 support

bence92:
* `force_text` was removed in favor of `force_str`
* `ugettext_lazy` was removed in favor of `gettext_lazy`
* added Python 3.9 and 3.10 to the test matrix, because they're supported by Django 4.0 (while excluding Django versions, which do not support 3.9 and 3.10)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,8 @@ jobs:
       fail-fast: false
 
       matrix:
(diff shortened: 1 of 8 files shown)


safwanrahman (approved):
Looks good to me. Thanks @bence92 🚀

sondrelg:
Do you intend on releasing a patch version with these changes soon @safwanrahman? I see it was merged a few days ago 🙂

PR #412, opened 2022-06-23
optimize index nested/object fields

alexander3771:
Hi!

We don't need to create brand new document for every field we index


alexander3771:
Hello, @safwanrahman! 
Please take a look, this is small fix just to increase speed of index nested objects

safwanrahman (approved):
LGTM. r+

PR #414, opened 2022-07-20
Celery signal processor implement

Bidaya0:
base on https://github.com/django-es/django-elasticsearch-dsl/pull/87


safwanrahman:
This needs careful review and run into local machine. Can you add some documentation?

Bidaya0:
> This needs careful review and run into local machine. Can you add some documentation?

ok i will update this branch next week

safwanrahman:
@Bidaya0 Hi, Can you add some documentations?

Bidaya0:
Thanks for the reminder, I'll add it

> @Bidaya0 Hi, Can you add some documentations?

Bidaya0:
Documentations have been added.
@safwanrahman

requirements.txt
```diff
@@ -1,3 +1,3 @@
 django>=1.9.6
 elasticsearch-dsl>=7.0.0,<8.0.0
-
+celery>=4.1.0
```
safwanrahman, line 3: It should not be required

django_elasticsearch_dsl/signals.py
```diff
@@ -96,6 +99,124 @@ def teardown(self):
...
+            """Handle save with a Celery task.
+
+            Given an individual model instance, update the object in the index.
+            Update the related objects either.
+            """
+            pk = instance.pk
+            app_label = instance._meta.app_label
+            model_name = instance.__class__.__name__
+
+            self.registry_update_task.delay(pk, app_label, model_name)
```
safwanrahman, line 130: It is better to use [`transaction.on_commit`](https://docs.djangoproject.com/en/4.2/topics/db/transactions/#django.db.transaction.on_commit) hook for this kind of thing. Otherwise, if the model is saved inside a atomic transaction, The task will not be able to fetch the object.

django_elasticsearch_dsl/signals.py
```diff
@@ -96,6 +99,124 @@ def teardown(self):
...
+            self.registry_update_related_task.delay(pk, app_label, model_name)
```
safwanrahman, line 131: This one as well. Need to add `on_commit` hook.

django_elasticsearch_dsl/signals.py
```diff
@@ -96,6 +99,124 @@ def teardown(self):
...
+                    related = doc_instance.get_instances_from_related(instance)
+                except ObjectDoesNotExist:
+                    related = None
+                if related is not None:
+                    doc_instance.update(related)
+                    if isinstance(related, models.Model):
+                        object_list = [related]
+                    else:
+                        object_list = related
+                    bulk_data = list(doc_instance._get_actions(object_list, action)),
```
safwanrahman, line 165: `_get_actions` is a private method. I am not sure, whether we should call this private method. Can you come up with any better idea?

django_elasticsearch_dsl/signals.py
Bidaya0: It has been changed, but it seems that on_commit is not triggered in django's TestCase, and I am still thinking of a suitable way to reuse test cases

django_elasticsearch_dsl/signals.py
Bidaya0: The public method corresponding to this method has been added

Bidaya0:
If you have time, review the revised code, plz.
@safwanrahman

Bidaya0:
@safwanrahman
Hello! Have time to check this out again?

mike-ivs:
Revisiting this topic 2years later and finding this has made my day! Very keen to see this PR integrated!

safwanrahman:
I have finally gone through the code and this looks good. Some minor adjustment need to be made, but I think it is good enough to be merged.
Thanks a lot @Bidaya0 , I am sorry that it took so long to get it merged.
🥳🤩
r+

safwanrahman (approved):

PR #444, opened 2023-03-17
Add mapping for PositiveBigIntegerField type

paulsmith:
This change allows ES to index Django model fields that are of the PositiveBigIntegerField type.

https://docs.djangoproject.com/en/4.1/ref/models/fields/#positivebigintegerfield


safwanrahman (approved):
LGTM. r+

PR #450, opened 2023-05-30
feat: Queryset can now be filtered when populating index.

qcoumes:
The idea of this MR is to allow filtering the queryset when populating the index.

Note that this only work by filtering models (with `--models`) which have the same column.

---

`filter` and `exclude` arguments have been added to the methods `DocType.get_queryset` and `DocType.get_indexing_queryset()`. They take a Django `Q` object (both default to `None`) and are applied to the queryset.

Similarly, parameters `--filter` and `--exclude` have been added to the `search_index` command.

These parameters take a list of filter in the form `<lookup>=<value>`. Almost all django lookups are supported, including one chaining another (e.g. `date_column__year__gt`)  and lookup spanning relationship (`e.g.manufacturer__name`).

The value can either be `None` (`nullable_column=`) an `int` (`integer_column=2`), a `float` (`float_column=2.1`), a `datetime` (`datetime_column=2020-12-20T12:34+02:00`) or a list of the previous types (`integer_column=2,3,4`).
If none of the types above matched, the value will be interpreted as a string.

BREAKING CHANGE: Indices which overrode `get_indexing_queryset` or `get_queryset` need to define the `filter=None` and `exlude=None` arguments.

### DIFF
--- a/django_elasticsearch_dsl/documents.py
+++ b/django_elasticsearch_dsl/documents.py
@@ -91,17 +91,24 @@ class DocType(DSLDocument):
             model=cls.django.model
         )
 
-    def get_queryset(self):
+    def get_queryset(self, filter=None, exclude=None):
         """
         Return the queryset that should be indexed by this doc type.
         """
-        return self.django.model._default_manager.all()
+        qs = self.django.model._default_manager.all()
 
-    def get_indexing_queryset(self):
+        if filter:
+            qs = qs.filter(filter)
+        if exclude:
+            qs = qs.exclude(exclude)
+
+        return qs
+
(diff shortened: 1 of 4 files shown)


django_elasticsearch_dsl/documents.py
```diff
@@ -91,17 +91,24 @@ def search(cls, using=None, index=None):
             model=cls.django.model
         )
 
-    def get_queryset(self):
+    def get_queryset(self, filter: models.Q = None, exclude: models.Q = None):
```
safwanrahman, line 94: We do not use type hint in this project.

qcoumes: Ok, I removed them 👍

PR #458, opened 2023-09-13
Added support for Elasticsearch 8

oehrlein:
I've made changes so Elasticsearch 8 works with this package.

The big caveat here is that testing via GitHub actions fail, but they run without error locally. If anyone could help get that fixed, that would be greatly appreciated as I don't have much experience with Github actions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,6 +37,8 @@ Features
 The library is compatible with all Elasticsearch versions since 5.x
 **but you have to use a matching major version:**
 
+- For Elasticsearch 8.0 and later, use the major version 8 (8.x.y) of the library.
+
 - For Elasticsearch 7.0 and later, use the major version 7 (7.x.y) of the library.
 
 - For Elasticsearch 6.0 and later, use the major version 6 (6.x.y) of the library.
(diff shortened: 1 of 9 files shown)


DarkAbhi:
> I've made changes so Elasticsearch 8 works with this package.
> 
> The big caveat here is that testing via GitHub actions fail, but they run without error locally. If anyone could help get that fixed, that would be greatly appreciated as I don't have much experience with Github actions.

The cause for the actions failing is because python2.7 support was removed from Actions some time around June.

safwanrahman:
Thanks @oehrlein for the PR. I need to test it locally with ES 8. Have you tested it locally?

oehrlein:
@DarkAbhi All the other Python versions are failing as well. I've toyed around with it for a bit and has to do with the security settings that come with Elasticsearch 8, namely the security certificate.

Regarding Python 2.7 not being supported by GitHub actions, I did see that as well, but didn't want to remove that in this PR since I think it'd be more fitting to remove it separately.

oehrlein:
@safwanrahman I have! All the tests pass for me.

tevariou:
@oehrlein @safwanrahman would it be possible to open a new branch for 7.x and merge this pr on main with https://github.com/oehrlein/django-elasticsearch-dsl/pull/6 ?

safwanrahman:
There are some fixes that need to be done in Github CI to pass the tests in CI. I think this PR is ready to go!
Thanks a lot @oehrlein. r+ from my end.

